PR #42, opened 2021-08-05
Add feature_details.md to clarify design and API details.

khushalsagar:


feature_details.md
```diff
@@ -0,0 +1,174 @@
...
+
+An important assumption in the design of this feature is that it's reasonable to limit customization options for cross-origin transitions to make it easier for UAs to enforce the constraints above. But the API/implementation should strive to not unnecessarily limit capabilities for same-origin transitions due to cross-origin constraints.
+
+
+## Design
+**Disclaimer** : The details in this design focus on the SPA/MPA use-case with extensive customizability. It’s expected that the functionality will be delivered incrementally with the initial version of the API being quite limited in scope. The purpose of exploring the customizability options is to ensure the API can evolve to accommodate them.
+
+There are 2 kinds of semantic transitions which are a part of this feature :
+
+* Enter/Exit Transition : An element which exists only on the new Document animates as it enters the screen. The final state of the element in the new Document provides an end state for the animatable properties. Similarly an element which exists only on the previous Document animates as it exits the screen, with a start state for the animatable properties.
```
vmpstr, line 39: Just a general comment, we could probably use a glossary document / section so that we can agree whether it's called previous/new Document or incoming/outgoing Document or something else (and same for other terms that may come up)

feature_details.md
```diff
@@ -0,0 +1,174 @@
...
+
+* Paired Transition : An element which exists on both Documents is automatically animated from the start state on the previous Document to the end state on the new Document. The painted content for these elements may or may not be identical, which requires defining the behaviour for transitioning this content.
+
+### Control Flow
+This section covers the flow of events during a Document transition for the MPA use-case. The SPA scenario is identical except the events may map to different API points and are dispatched on the same Document. The code-snippets used are a rough API sketch with detailed explanation in subsequent sections.
+
+* The entry point for this API is document.documentTransition. The previous Document provides a url keyed list of elements, referred to as shared elements, which could be animated during a transition. There are a couple of important assumptions here :
```
vmpstr, line 46:
```suggestion
* The entry point for this API is a new `document.documentTransition` object. The previous Document provides a url keyed list of elements, referred to as shared elements, which could be animated during a transition. There are a couple of important assumptions here :
```

vmpstr, line 46: What is a url keyed list of elements? (probably a good addition for glossary section)

feature_details.md
```diff
@@ -0,0 +1,174 @@
...
+
+* When a navigation is initiated to this url, the UA caches a static pixel snapshot for the DOM sub-tree of each shared element. If there is any hierarchical relationship between these elements, they are captured as separate snapshots to allow them to animate independently. The parent element's snapshot excludes content for any embedded shared elements, drawing the background which was earlier occluded by the child element.
+
+* The feature supports a subset of [animatable properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_animated_properties) that can be interpolated during a transition. For each of those properties, the UA caches their start value to animate them during the transition. In addition to this metadata, we allow the developer to provide opaque contextual information (a unique identifier for the element, navigation trigger, scroll offset, site version etc.), referred to as propertyMap, about these elements to configure the transition.
+
+```
+document.documentTransition.setSameOriginTransitions(
```
vmpstr, line 57: Can you please format this a bit more verbosely?

feature_details.md
```diff
@@ -0,0 +1,174 @@
...
+  {“foo.com/a.html” : [{element : document.html, propertyMap : {“id” : “root”}},
+                                 {element : document.getElementById(‘shared’), propertyMap : {“id” : “shared”}],
+   “foo.com/b.html” : [{element : document.html}]
+  });
+```
```
vmpstr, line 62: Can you add an explanation for what the above code sample is meant to do

feature_details.md
```diff
@@ -0,0 +1,174 @@
...
+
+* When the new Document loads, the developer can register an `EventListener` to be notified if a transition was initiated by the previous Document. The event, referred to as `handleTransition`, provides the metadata for elements captured from the previous Document and the previous URL, a detailed explanation is at [Referencing Elements](#referencing-elements). The aim of the `handleTransition` event is to allow the developer to define the animations for a transition using information from both Documents. For instance, a paired element transition may be dropped if the corresponding element in the new Document renders offscreen.
```
vmpstr, line 64: > new Document loads

Do you mean the onload event fires? I think that may be too late to start a transition.

feature_details.md
```diff
@@ -0,0 +1,174 @@
...
+* The `handleTransition` event provides a promise based API to declare the transition parameters asynchronously. The motivation to make this asynchronous is to allow the new Document to go through intermediate layouts before it is ready to be displayed. The first paint for the Document is deferred and the previous Document is displayed until this promise is fulfilled and the transition can be started.
+
+```
+document.documentTransition.addEventListener(“handleTransition”, e => {
+    e.initializeTransition(transitionFrom(e.url, e.elements));
+});
+
+async function transitionFrom(previousUrl, elements) {
+    // Wait for resources required for this transition to load.
+    await waitForLoad(previousUrl);
```
vmpstr, line 75: What is this waiting for? Does that mean that while we're running script for this page, we're still executing visual presentation from the previous page?

feature_details.md
```diff
@@ -0,0 +1,174 @@
...
+
+    elementTransitions = []
+    for (let elem in elements) {
+        if (elem.propertyMap[“id”] === “root”) {
+          elementTransitions.push({sourceElement : elem, transitionType: “none”});
```
vmpstr, line 80: Can you add some comments in this code snippet, since it's not very clear what this does

feature_details.md
```diff
@@ -0,0 +1,174 @@
...
+          elementTransitions.push({destinationElement : document.html, transitionType : “fade”});
+        } else if (elem.propertyMap[“id”] === “shared”) {
+          elementTransitions.push({sourceElement : elem, destinationElement : document.getElementById(“shared”));
+        }
+    }
+    return elementTransitions;
+}
+```
+
+* The first rendering lifecycle update after the promise returned by the API resolves is when the animations for this transition are started. This frame also determines the end value for animated properties of shared elements in the new Document, similar to the step on the previous Document to cache the start value for its elements.
```
vmpstr, line 90: > after the promise returned by the API resolves

not sure which API call you mean here

feature_details.md
```diff
@@ -0,0 +1,174 @@
...
+
+* The developer can register another event listener to be notified when all animations associated with this transition have been finished. A caveat to note here is that since the transition is executed with a live DOM for the new Document, any changes to the shared elements in the new Document before this event is dispatched will have undefined behaviour.
```
vmpstr, line 92: > any changes to the shared elements in the new Document before this event is dispatched will have undefined behaviour.

That doesn't seem ideal. Do you mean that if a shared element has, for example, an animated position via css animations then we can't use that element as a shared target? Why not?

feature_details.md
```diff
@@ -0,0 +1,174 @@
...
+
+* The container size for the element animates from the paint bounds of the previous element to the paint bounds of the new element. During this animation, neither element’s content fits the size of the container. Fitting the painted content within the container is similar to a resize operation to fit a [replaced element’s](https://developer.mozilla.org/en-US/docs/Web/CSS/Replaced_element) content. The default animation uses the behaviour defined by fill mode in [object-fit](https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit) but customizing this should be supported going forward. TODO : Add example.
+
+* The pixels from the painted content of the 2 elements are blended using a [cross-fade](https://developer.mozilla.org/en-US/docs/Web/CSS/cross-fade()) animation from `cross-fade(previous 100%, new 0%)` to `cross-fade(previous 0%, new 100%)`.
```
vmpstr, line 114: I'm unsure about the level of detail in these descriptions. This sounds like we must use the cross-fade algorithm as specified. If the algorithm is changed, then our implementation should change as well. Is this the intent of this text?

The problem is that the blending would typically happen in a different code location / rendering pipeline, so I'm not sure how faithfully we can implement the exact cross-fade algorithm (or rather how much it matters)

feature_details.md
```diff
@@ -0,0 +1,174 @@
...
+
+It’s conceivable that support for animatable properties will be added incrementally and not all will be supported across UAs. This is still an open question and one option could be for the API to allow feature detection to detect which properties are supported by the UA. Using an unsupported property on a shared element has undefined behaviour.
+
+#### Animation Configuration
+The following fields are used in a dictionary, referred to as `ElementTransition`, to configure a single enter/exit or paired transition :
+
+* `transitionType` : This is an enum to provide a predefined set of animation patterns for enter/exit transitions. These can be explicit patterns like `slide-left` (does a translation in the specified direction) which provide convenient defaults for common use-cases. For instance, sliding a header/footer offscreen. Or higher level abstractions like `previous`/`next` which allow the UA to decide the appropriate animation (for example based on the idiomatic pattern for the native OS).
+
+* `duration` : The total length for this transition.
```
vmpstr, line 123: Is this an arbitrary duration or is it capped by some value?

feature_details.md
```diff
@@ -0,0 +1,174 @@
...
+
+* `delay` : A Document transition establishes a universal timeline for animations spanning multiple elements. This field indicates the delay in starting animations for this transition on this universal timeline.
+
+* `element` : A reference to the [Element](https://dom.spec.whatwg.org/#interface-element) in the new Document.
+
+* `sharedElement` : A reference to an element in the previous Document, explained in Referencing Elements (#referencing-elements).
+
+If one of `sharedElement` or `element` is set, this is an enter or exit transition respectively. The animations for each animatable property are defined based on `transitionType`. If both `sharedElement` and `element` are set, this is a paired transition. This transition first creates a combined image using painted content from both elements based on the `content` animation specification under [Animatable Properties](#animatable-properties). Other animatable properties are then applied to this combined output.
```
vmpstr, line 131:
```suggestion
If exactly one of `sharedElement` or `element` is set, this is an enter or exit transition respectively. The animations for each animatable property are defined based on `transitionType`. If both `sharedElement` and `element` are set, this is a paired transition. This transition first creates a combined image using painted content from both elements based on the `content` animation specification under [Animatable Properties](#animatable-properties). Other animatable properties are then applied to this combined output.
```

vmpstr, line 131: >  This transition first creates a combined image using painted content from both elements

I'm not sure what this means. If we have a border-radius 'from' element, and a opacity 0.5 but no border radius 'to' element, then what is the 'combined image' as you describe it here?

feature_details.md
```diff
@@ -0,0 +1,174 @@
...
+
+An [open request](https://github.com/WICG/shared-element-transitions/issues/28) here is the ability to add low-level customization to these animations by specifying a delay, duration, easing curve, keyframes similar to the [Web Animation API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Animations_API). A potential option to provide this capability in future iterations is to add a `keyframes` field to this dictionary which supports property types listed under [Animatable Properties](#animatable-properties). The syntax is defined under [Processing a keyframes argument](https://drafts.csswg.org/web-animations-1/#processing-a-keyframes-argument) in Web Animations API. It’s worth mentioning that the keyframes syntax already has a concept of an implicit start/end keyframe which could map to the start/end state for each property computed in the previous/new Document respectively.
```
vmpstr, line 133: > A potential option to provide this capability in future iterations is to add a `keyframes` field to this dictionary which supports property types

This seems like we would have to reimplement web animations to a degree, just in a different spot. Is this really the request?

feature_details.md
```diff
@@ -0,0 +1,174 @@
...
+
+#### Referencing Elements
+Specifying a transition above requires the new Document to reference elements from the previous Document. We address this problem, along with flexibility to exchange contextual information between Documents, by introducing a placeholder for shared elements captured from the previous Document. The documentTransition APIs mentioned in the following explanation are clarified under [Control Flow](#control-flow).
+
+* The dictionary used to specify a shared element cached from the previous Document in documentTransition.setSameOriginTransitions has the following fields :
+    * `element` : A reference to the [Element](https://dom.spec.whatwg.org/#interface-element) in the previous Document.
```
vmpstr, line 139: Is this an actual Element that you can set properties to? I thought this would be more akin to an opaque handle, rather than an Element object, especially since it's referencing a document that is no longer active

feature_details.md
```diff
@@ -0,0 +1,174 @@
...
+    * `propertyMap` : A `record<DOMString, any>` provided by the developer to pass any opaque contextual information to the new Document. The value for this map can be any javascript object supported by the [structured clone algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm).
+    * `metadataOnly` : A boolean indicating whether the UA should skip capturing the element snapshot, defaults to false. This is purely a performance optimization when a visually identical element is present in both Documents, which would be a common scenario for paired transitions.
```
vmpstr, line 141: I'm not sure about the order of events here. I thought this API is happening on the incoming page _after_ the visual contents of the outgoing element have been captured. So, is this specified by the incoming page or the outgoing? In either case, I think this API is complicated enough that we can leave this off for now, especially if it's only an optimization. We can certainly add it later

feature_details.md
```diff
@@ -0,0 +1,174 @@
...
+
+* Paired transitions are executed by creating a combined image for elements in the two trees on which the animated properties are applied. Should the transition switch to using a parent stacking context for the shared element in the new Document when a transition starts?
+
+The following steps define the behaviour of how content from the two Documents is drawn together during the transition. The aim is to avoid an element’s visual representation abruptly changing at transition start/end for common use-cases :
+
+* We assume that the dimensions for the initial containing block for both Documents is the same. If this changes, for example because the user resizes the browser window before the transition finishes, the transition is aborted.
+
+* During the transition, the UA creates a new root element (hidden from script), referred to as `TransitionRoot` which establishes the root stacking context and containing block. The background color for the `TransitionRoot` is the same as the background color for the root element in the new Document. By default, all shared elements (including root elements from both Documents) are direct descendants of the `TransitionRoot`.
```
vmpstr, line 156: This seems to be describing a particular implementation, rather than effect. Or do you mean to say that this is an example of how it might be implemented?

feature_details.md
```diff
@@ -0,0 +1,174 @@
...
+
+* TODO : Clarify the details of how properties are computed and animated during the transition.
+
+* We also need to define the order in which elements which share a stacking context are rendered. The current proposal allows the developer to explicitly control this by providing an ordered list (back-to-front) of `ElementTransition` in the `handleTransition.initiateTransition` API. But it’s unclear if this control is unnecessary complexity. The common use-case would be to maintain the same visual order for elements within the previous and new Document to avoid abrupt changes when the transition starts and ends.
```
vmpstr, line 160: I keep coming back to the fact that this is a potentially 300ms transition where some buttons and logos fly around. I agree that the relative paint order of element is something that needs to be defined -- perhaps it can inferred from the relative paint order of either the incoming or outgoing pages -- but I'm against having this control in developer's hands, because at the end of the day I don't think it matters and it only complicates the API. However, I'm happy to be proven wrong here.

feature_details.md
```diff
@@ -0,0 +1,174 @@
+# Feature Details
+
+This document is a detailed description of this feature along with the design assumptions, constraints and targeted use-cases. The aim is to facilitate API discussion and ensure it's structured with a focus on interoperability.
+
+Note that since the API details are still in flux, this may not be consistent with other documentation. Please refer to the documentation in the README file and sample-code to experiment with the current implementation in Chromium.
```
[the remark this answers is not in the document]
khushalsagar: I started another pull request to reflect the API change which was missing in the README file : https://github.com/WICG/shared-element-transitions/pull/43

feature_details.md
khushalsagar: Good point. Let's stick with incoming/outgoing for now and I'll clarify this in a Glossary Section.

feature_details.md
khushalsagar: This is a detail that still needs to be clarified. The exact stage in the navigation's lifecycle when this event is dispatched.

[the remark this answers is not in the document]
khushalsagar: Agreed about being careful about the timing. But same as the comment above, the exact timing is an open question.

feature_details.md
khushalsagar: I'll add the explanation for this with comments in the doc itself but here is a quick clarification.

This is waiting for any async activity required for first paint of the new Document to finish. The most common would be fetching assets from the network (images, fonts, scripts). The developer can decide how long to wait before the transition is started and what the fallback should be if we hit a timeout. For example, if an image on the new Document couldn't be loaded. What should we do if this image was a part of a paired transition?

feature_details.md
khushalsagar: Thanks for this question!

The main intent here is that the blending function used to combine the 2 images must ensure that blending 2 pixels with the same value is a no-op, or B(a, a) = a. CSS cross-fade's  implementation already does so I was trying to build on the existing concept of how to blend 2 images. I looked at its spec now : https://drafts.csswg.org/css-images-4/Overview.bs, and even that doesn't describe the exact blending operation the implementation must use.

It sounds reasonable to clarify the intent here for now and we can decide the level of detail that must be addressed here based on spec discussions?

feature_details.md
khushalsagar: This is indeed the request, yes.

The idea of lower level customization to control the following aspects sounds totally reasonable to me :
- Easing curve for each property, a separate value for opacity vs transform for example.
- Delay to stagger the animations. For example the opacity animation starts after the transform animation has finished.
- Separate duration for each property.

But it's still worth discussing whether specifying values for each property is built on top of CSS transitions (animating between 2 states) or keyframes as specified in Web Animations. I don't think we need to resolve this question at this stage, but we should ensure the API can be extended to accomodate this request.

feature_details.md
khushalsagar: Hmmm, I wasn't trying to push for a specific implementation. My aim here was to explain how elements are drawn during a transition using web platform concepts so we can understand the space in which properties like 'opacity' or 'transform' exist.

For example, the transition could animate the transform for the root element of a Document. In the default case this transform would apply to its entire subtree. But during the transition, the shared elements from this Document are "detached" from their root element. The concept of a temporary "root" element as described here was a way to explain the relationship between a Document's root element and its shared elements during the transition.

feature_details.md
khushalsagar: I agree that this doesn't need to be a developer control but we'll need to think through a few cases to define the behaviour here : 

a) The order in which root elements draw, old or new on top. The current implementation uses a single enum to capture that behaviour. But in this proposal I was trying to define them as 2 single transition types : {new:none, old: slide-left}, where old  content draws on top and slides left to reveal the new content.

b) Assuming shared elements within a document draw in the same order. Is it all old elements stacked behind new ones or vice versa.

c) What do we do for paired elements.

khushalsagar:
A new change is ready for review!

feature_details.md
```diff
@@ -0,0 +1,417 @@
...
+
+## Sample Code
+This section explains how the feature can be used to add transitions to MPAs. There are a set of use-cases and edge cases to consider :
+
+### Root Transition
+Let's start with a root transition, which adds an animation to the root element of previous and new Documents. Below is a video of the transition we want to create :
+
+[![Video Link for Root Element Transition](https://img.youtube.com/vi/mCvvWMIGItY/0.jpg)](https://www.youtube.com/watch?v=mCvvWMIGItY)
+
+The entry point for this API is a new `document.documentTransition` object. The previous Document provides a list of elements which will be animated during the transition. Since this is root transition, this list only includes the `html` element. The result of calling `setSameOriginTransitionElements` below is that the browser caches a copy of the pixels (snapshot) currently present on the screen when a navigation to `foo.com/a.html` is initiated.
```
sumitbhagwani, line 50: The result of calling `documentTransition.foo` API/method as illustrated below

sumitbhagwani: is it because only element in list is the `html` element?

feature_details.md
```diff
@@ -0,0 +1,417 @@
...
+  {
+   “foo.com/a.html” : [{element : document.html}],
+  });
+```
+
+When the new Document loads, script registers an event listener to be notified if a transition was initiated by the previous Document. The exact Document lifecycle stage before which this event should be registered is TBD.
+
+```
+document.documentTransition.addEventListener(
+    “handleTransition”, e => {
```
sumitbhagwani, line 66: can you mention that this is a keyword somewhere

feature_details.md
```diff
@@ -0,0 +1,417 @@
...
+
+And specify a transition which associates these elements in the `handleTransition` event on the new Document :
+
+```
+document.documentTransition.addEventListener(
+    “handleTransition”, e => {
+        e.startTransition(transitionFrom(e.previousUrl, e.previousElements));
+    });
+
+async function transitionFrom(previousUrl, previousElements) {
```
sumitbhagwani, line 157: `provideTransitionsFrom` would it make it clearer that this method returns transitions.

feature_details.md
```diff
@@ -0,0 +1,417 @@
...
+    // Specify an animation for the old image to exit revealing new
+    // image.
+    let imageAnimation = 
+      {previousElement : previousElements[1],
+       transitionType : type};
+    return [imageAnimation];
+}
+```
+
+### Integration with App History API
```
sumitbhagwani, line 271: add a link to this API explainer

feature_details.md
```diff
@@ -0,0 +1,417 @@
...
+* The animations associated with the transition need to be defined completely by the previous Document, which is aware of the site the user is navigating to and has contextual information about the user journey.
+
+* The animations need to be executed such that there are no observable side-effects for the new and previous Documents. For example, setting an obscure animation for a specific element on the incoming Document could be used to exchange information between origins. Enforcing no side-effects is an important consideration, since it can influence the stage in the rendering pipeline where the UA executes these animations.
+
+An important assumption in the design of this feature is that it's reasonable to limit customization options for cross-origin transitions to make it easier for UAs to enforce the constraints above. But the API/implementation should strive to not unnecessarily limit capabilities for same-origin transitions due to cross-origin constraints.
+
+## Sample Code
+This section explains how the feature can be used to add transitions to MPAs. There are a set of use-cases and edge cases to consider :
+
+### Root Transition
```
sumitbhagwani, line 45: it'll be nice to code snippet and explanation side by side.

khushalsagar: Finding it hard to put a mult-line code snippet next to a description with a table.

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+* The animations associated with the transition need to be defined completely by the previous Document, which is aware of the site the user is navigating to and has contextual information about the user journey.
+
+* The animations need to be executed such that there are no observable side-effects for the new and previous Documents. For example, setting an obscure animation for a specific element on the incoming Document could be used to exchange information between origins. Enforcing no side-effects is an important consideration, since it can influence the stage in the rendering pipeline where the UA executes these animations.
+
+An important assumption in the design of this feature is that it's reasonable to limit customization options for cross-origin transitions to make it easier for UAs to enforce the constraints above. But the API/implementation should strive to not unnecessarily limit capabilities for same-origin transitions due to cross-origin constraints.
+
+## Glossary
+| Term | Description |
+| ------------- | ------------- |
+| Previous Document  | The Document the user is viewing when a navigation is initated.  |
```
vmpstr, line 45: It'd be nice to mention what this means in SPA cases as well

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+
+The entry point for this API is a new `document.documentTransition` object. The previous Document provides a list of elements which will be animated during the transition. Since this is root transition, this list only includes the `html` element. The result of calling `documentTransition.setSameOriginTransitionElements` as illustrated below is that the browser caches a copy of the pixels (snapshot) for the `html` element when a navigation to `foo.com/a.html` is initiated.
+
+```
+// The list of elements is keyed on url. When a user navigates to
+// a url, the corresponding entry in this dictionary is used to
+// decide which elements should be cached.
+document.documentTransition.setSameOriginTransitionElements(
+  {
+   “foo.com/a.html” : [{element : document.html}],
```
vmpstr, line 68: I don't quite understand why you need to specify the url, including the domain, for a same origin transition. Also, I missed that this was for MPA and was even more confused :P Maybe mentioning again somewhere that this is for MPA specifically would be nice.

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+  });
+```
+
+Now we need to add some code on the new Document. When the new Document loads, script registers a listener for a new event type called `handleTransition`. The event is dispatched if a transition was initiated by the previous Document. The exact Document lifecycle stage before which this event should be registered is TBD.
+
+```
+document.documentTransition.addEventListener(
+    “handleTransition”, e => {
```
vmpstr, line 76: One thing we need to worry about is the event firing before the event handler is registered. For example, if we fire the event after the above-the-fold content has been parsed and presented, and the event handler is at the end of the page that hasn't been parsed yet, then we will miss the event.

Nothing to change in the doc here, but just something to keep in mind.

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+        e.startTransition(provideTransitionFrom(e.previousUrl, e.previousElements));
+    });
+
+async function provideTransitionFrom(previousUrl, previousElements) {
+    // Wait for the image on the new page to load before
+    // starting the transition.
+    await waitForImageLoad();
```
vmpstr, line 83: In order for this code snippet to do what I expect it wants to do, the page has to still be displaying the previous page's content while this line (and the image itself) is loading. I think it may be possible to delay presenting new content in the same way Chromium does with paint holding, but I'm just curious if we intend for this API to block presentation of new content even further (and for arbitrarily long).

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+
+Adding an event listener for `handleTransition` will result in execution of following steps :
+
+* The browser keeps displaying content for the previous Document when `handleTransition` is dispatched. This does not require the previous Document to be active since the browser can display a cached pixel copy.
+
+* The `handleTransition` event has a field `previousURL` : the URL for the previous Document. And `previousElements` : the list of placeholders for elements passed to `setSameOriginTransitionElements`, clarified in [API Proposal](#api-proposal).
+
+* The `handleTransition` event has an API `startTransition` which takes the list of elements to animate and the type of animation. In the example above, `provideTransitionFrom` returns the root elements for the previous and new Document. Also note that the animation done is based on `transitionType` specified with each element. This is a new enum with a pre-defined list of animation patterns.
+
+* The `startTransition` API takes a promise to allow script to asynchronously prepare the new Document for first paint. The browser continues to display old content until the promise passed to this API resolves.
```
vmpstr, line 108: The last two bullet points seem to contradict each other: does `startTransition` take a list of elements to animate or does it take a promise?

feature_details.md
khushalsagar: This needs to be refined more, I don't expect the eventual API would require exact URL strings. Especially because a common scenario would be to use regex pattern to specify the same set of elements for a set of pages. I'll add a comment clarifying that this needs more work. We can probably take inspiration from pre-render speculation rules for structuring this : https://github.com/jeremyroman/alternate-loading-modes/blob/main/triggers.md#speculation-rules

But the goal of the API is to allow the author to configure sharing a different list of elements for any same origin navigation.

feature_details.md
khushalsagar: Noted. With respect to timing for this event, I assumed the trickier problem is to identify the stage in the navigation lifecycle when it is appropriate to dispatch this event. Some notes from digging right now, it needs to be between step 5/6 in the history traversal algorithm : https://html.spec.whatwg.org/multipage/browsing-the-web.html#history-traversal. I expect this needs to be fired between the document.readyState changing from loading to interactive : https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState

I didn't realize the position in the page where tha handler is registered could also be a difficult aspect for the developer to reason about. Thanks for pointing this out!

feature_details.md
khushalsagar: The intent indeed is to use paint holding to allow the API to be used to block presentation of new content. I think it's reasonable for the browser to have a timeout (currently we use 4 seconds for the SPA) and how long that should be is up for discussion. But my thinking is that the way an SPA can completely control the experience of delaying transitions between documents, an MPA should be able to do it as well.

feature_details.md
khushalsagar: Done.

feature_details.md
khushalsagar: Added a comment to clarify this in the API proposal section.

feature_details.md
khushalsagar: Added a TODO in API proposal to make a note of this.

feature_details.md
khushalsagar: Fixed the wording. Thanks!

vmpstr (approved):

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+});
+```
+
+Special handling is needed for a list of elements : `[CacheElement_Parent, CacheElement_Child]` where `CacheElement_Child` is a descendant of `CacheElement_Parent`. The parent element's snapshot excludes content for any descendant elements. The space covered by the descendant element shows the background which was earlier occluded by the descendant element.
+
+*TODO: Clarify the syntax for nextURL. This should allow for [URL patterns](https://github.com/WICG/urlpattern).*
+
+### New Document
+#### Referencing Elements
+The first concept for this feature on the new Document is a class : `PreviousDocumentElement`. This class provides a placeholder used by script in the new Document to reference elements from the previous Document. A `PreviousDocumentElement` object in the new Document maps 1:1 to a `CacheElement` object in the previous Document. This class has a `getPropertyMap()` API to retreive the `record<DOMString, any>` for the corresponding `CacheElement`.
```
jakearchibald, line 320:
```suggestion
The first concept for this feature on the new Document is a class: `PreviousDocumentElement`. This class provides a placeholder used by script in the new Document to reference elements from the previous Document. A `PreviousDocumentElement` object in the new Document maps 1:1 to a `CacheElement` object in the previous Document. This class has a `getPropertyMap()` API to retreive the `record<DOMString, any>` for the corresponding `CacheElement`.
```

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+
+#### Animation Configuration
+The Shared Element Transition feature allows the web author to compose a transition by animating a set of elements. For example, animating the roots and sub-elements in [Single Element Transition](#single-element-transition) and [Paired Element Transition](#paired-element-transition). We divide these animations into the following semantic transition types :
```
jakearchibald, line 323:
```suggestion
The Shared Element Transition feature allows the web author to compose a transition by animating a set of elements. For example, animating the roots and sub-elements in [Single Element Transition](#single-element-transition) and [Paired Element Transition](#paired-element-transition). We divide these animations into the following semantic transition types:
```

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+
+* Single Element Transition : An element which exists only on the new Document animates as it enters the screen. An element which exists only on the previous Document animates as it exits the screen. A Root transition can be built using two Single Element Transitions.
```
jakearchibald, line 325:
```suggestion
* Single Element Transition: An element which exists only on the new Document animates as it enters the screen. An element which exists only on the previous Document animates as it exits the screen. A Root transition can be built using two Single Element Transitions.
```

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+
+* Paired Element Transition : An element which exists on both Documents is automatically animated between them. These elements may or may not be visually identical.
```
jakearchibald, line 327:
```suggestion
* Paired Element Transition: An element which exists on both Documents is automatically animated between them. These elements may or may not be visually identical.
```

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+
+The parameters to confugure an instance of Single Element or Paired Element Transition are specified using a new dictionary type : `ElementTransition`. This dictionary has the following fields :
```
jakearchibald, line 329:
```suggestion
The parameters to configure an instance of Single Element or Paired Element Transition are specified using a new dictionary type: `ElementTransition`. This dictionary has the following fields:
```

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+
+* `transitionType` : This is an enum to provide a predefined set of animation patterns for Single Element transitions. These can be explicit patterns like `slide-left` (does a translation in the specified direction) or higher level abstractions like `previous`/`next` which allow the UA to decide the appropriate animation (for example based on the idiomatic pattern for the native OS). The exact types are TBD.
```
jakearchibald, line 331:
```suggestion
* `transitionType`: This is an enum to provide a predefined set of animation patterns for Single Element transitions. These can be explicit patterns like `slide-left` (does a translation in the specified direction) or higher level abstractions like `previous`/`next` which allow the UA to decide the appropriate animation (for example based on the idiomatic pattern for the native OS). The exact types are TBD.
```

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+
+* `duration` : The total length for this transition. The browser may cap this to a reasonable limit.
```
jakearchibald, line 333:
```suggestion
* `duration`: The total length for this transition. The browser may cap this to a reasonable limit.
```

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+
+* `delay` : A Document transition establishes a universal timeline for multiple `ElementTransition`s. This field indicates the delay in starting animations for this transition on this universal timeline.
```
jakearchibald, line 335:
```suggestion
* `delay`: A Document transition establishes a universal timeline for multiple `ElementTransition`s. This field indicates the delay in starting animations for this transition on this universal timeline.
```

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+
+* `newElement` : A reference to the [Element](https://dom.spec.whatwg.org/#interface-element) in the new Document.
```
jakearchibald, line 337:
```suggestion
* `newElement`: A reference to the [Element](https://dom.spec.whatwg.org/#interface-element) in the new Document.
```

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+
+* `previousElement` : A reference to a `PreviousDocumentElement` provided in the `handleTransition` event (clarified below).
```
jakearchibald, line 339:
```suggestion
* `previousElement`: A reference to a `PreviousDocumentElement` provided in the `handleTransition` event (clarified below).
```

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+
+### Transition Lifecycle
+The main entry point for the API in the new Document is the `handleTransition` [event](https://dom.spec.whatwg.org/#interface-event). This event is dispatched prior to first paint of a new Document if the previous Document had initialized a transition using `setSameOriginTransitionElements`. *TODO : Should this survive across redirects if the final URL is same origin?* Script can register for this event using `document.documentTransition` which is an [Event Target](https://dom.spec.whatwg.org/#interface-eventtarget). *TODO : Figure out the exact timing of this event in the Document's lifecycle. We also need to ensure that we allow script to register the event before first paint of the new Document.*
+
+The `handleTransition` event has the following fields :
+
+* `previousUrl` : The URL for the previous Document which initiated this transition.
```
jakearchibald, line 348:
```suggestion
* `previousUrl`: The URL for the previous Document which initiated this transition.
```

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+
+* `previousElements` : A list of `PreviousDocumentElement`s. These map 1:1 with the list of `CacheElement` passed to `setSameOriginTransitionElements`.
```
jakearchibald, line 350:
```suggestion
* `previousElements`: A list of `PreviousDocumentElement`s. These map 1:1 with the list of `CacheElement` passed to `setSameOriginTransitionElements`.
```

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+
+This event also has an API to start the transition : `void startTransition(Promise<sequence<ElementTransition>>)`. The precise semantics of this API are described below :
```
jakearchibald, line 352:
```suggestion
This event also has an API to start the transition: `void startTransition(Promise<sequence<ElementTransition>>)`. The precise semantics of this API are described below:
```

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+    await waitForNextScene();
+    
+    let oldRootAnimation = {previousElement : previousElements[0],
+                            transitionType : "slide-fade-left"};
+    let newRootAnimation = {newElement : document.html,
+                            transitionType : "none"};
+    
+    return [oldRootAnimation, newRootAnimation];
+}
+```
```
jakearchibald, line 384:
````suggestion
```js
async function doRootTransition() {
  await document.documentTransition.performSameDocumentTransition({
    cacheElements: [{ element: document.html }],
    startCallback: startTransition,
  });
}

async function startTransition(previousElements) {
  await waitForNextScene();

  let oldRootAnimation = {
    previousElement: previousElements[0],
    transitionType: "slide-fade-left",
  };
  let newRootAnimation = { newElement: document.html, transitionType: "none" };

  return [oldRootAnimation, newRootAnimation];
}
```
````

File: feature_details.md [hunk 43]
```diff
@@ -0,0 +1,413 @@
...
+
+
+## Security Considerations
+Since the design above is limited to same-origin transitions, the only security constraint is to ensure that script can never read pixel content for element snapshots. This is necessary since the Document may embed cross-origin content (iframes, CORS resources, etc.) and multiple restricted user information (visited links history, dictionary used for spell check, etc.)
+
+## Future Work
+Subsequent designs should cover the cross-origin use-case and address design considerations specific to that use-case. A few problems restricted to it are :
+
+* Opt-in vs Opt-out : The transition must be decided completely by the previous origin initiating the navigation. This necessitates the ability for the new origin to decide which origins can control a transition to it. It’s unclear whether this should be an opt-in or opt-out. The decision will need to tradeoff limiting the customizability of the transition to ensure there is no potential for abuse to allow an opt-out approach.
```
jakearchibald, line 409:
```suggestion
* Opt-in vs Opt-out: The transition must be decided completely by the previous origin initiating the navigation. This necessitates the ability for the new origin to decide which origins can control a transition to it. It’s unclear whether this should be an opt-in or opt-out. The decision will need to tradeoff limiting the customizability of the transition to ensure there is no potential for abuse to allow an opt-out approach.
```

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+
+* Referencing elements : Same-origin relies on sharing information across documents to reference elements between them, which won’t be an option for cross-origin transitions. This should leverage existing work for [scroll-to-css-selector](https://github.com/bryanmcquade/scroll-to-css-selector).
```
jakearchibald, line 411:
```suggestion
* Referencing elements: Same-origin relies on sharing information across documents to reference elements between them, which won’t be an option for cross-origin transitions. This should leverage existing work for [scroll-to-css-selector](https://github.com/bryanmcquade/scroll-to-css-selector).
```

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+
+* Time to First Paint : Same-origin relies on the new Document using an explicit signal to indicate when it is ready for display, which won’t be an option for cross-origin transitions. This will likely need to rely on UA heuristics to decide when the transition is started, especially with an opt-out approach.
```
jakearchibald, line 413:
```suggestion
* Time to First Paint: Same-origin relies on the new Document using an explicit signal to indicate when it is ready for display, which won’t be an option for cross-origin transitions. This will likely need to rely on UA heuristics to decide when the transition is started, especially with an opt-out approach.
```

jakearchibald (changes requested):
Mostly formatting nits for now. I'd like to give the API some more thought, but it seems like it's heading in the right direction.

jakearchibald:
@khushalsagar ah, I missed the deadline with my review. Can these fixes be used? I corrected a few errors in examples, plus a bunch of formatting.

feature_details.md
khushalsagar: Done.

feature_details.md
khushalsagar: Done.

feature_details.md
khushalsagar: Done.

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+
+* Both cross-document and same-document transitions require deferring first paint for the new scene which is controlled by script. It's unclear what or if this should have an upper bound.
+
+* Since transitions execute with a live new Document, the state of new Elements can change while a transition is in progress. For example, an image element used in the transition receives more data or is detached from the DOM. It's unclear whether the behaviour here needs to be defined.
+
+
+## Security Considerations
+Since the design above is limited to same-origin transitions, the only security constraint is to ensure that script can never read pixel content for element snapshots. This is necessary since the Document may embed cross-origin content (iframes, CORS resources, etc.) and multiple restricted user information (visited links history, dictionary used for spell check, etc.)
+
+## Future Work
```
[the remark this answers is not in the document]
khushalsagar: Done.

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+## Open Questions
+
+* The type of navigations the previous Document will be allowed to define a transition for should be similar to the ones supported by app-history’s [navigate event](https://github.com/WICG/app-history#restrictions-on-firing-canceling-and-responding). The most likely exception will be back/forward buttons which currently do not fire a `navigate` event. So a common pattern would be to use this API in response to the navigate event.
+
+* Both cross-document and same-document transitions require deferring first paint for the new scene which is controlled by script. It's unclear what or if this should have an upper bound.
+
+* Since transitions execute with a live new Document, the state of new Elements can change while a transition is in progress. For example, an image element used in the transition receives more data or is detached from the DOM. It's unclear whether the behaviour here needs to be defined.
+
+
+## Security Considerations
```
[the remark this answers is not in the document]
khushalsagar: Done.

feature_details.md
khushalsagar: Done.

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+* The browser keeps displaying visuals for the previous Document until the `handleTransition` event is dispatched. If no call to `startTransition` is made by an [Event Listener](https://dom.spec.whatwg.org/#callbackdef-eventlistener), the transition is aborted.
+
+* The `startTransition` API takes a list of `ElementTransition`s. These transition objects specify the set of animations to execute for the transition. We use a promise based API here to allow the new Document to asynchronously prepare the new Document. Since a page load frequently involves activity like network fetches or disk reads, allowing the prepare to be asynchronous is important.
+
+* The browser continues to display visuals for the previous Document until the promise provided in `startTransition` resolves. The first rendering lifecycle update after this step marks the new Document's first paint and transition start.
+
+We also provide a `transitionEnd` event to be notified when all animations associated with a transition have finished executing. There should ideall be no visual changes to DOM state between the transition start step above and `transitionEnd` event.
+
+
+### Single Page App API
```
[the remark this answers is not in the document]
khushalsagar: Done.

feature_details.md
khushalsagar: Done.

feature_details.md
khushalsagar: Done.

feature_details.md
khushalsagar: Done.

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+
+* `delay` : A Document transition establishes a universal timeline for multiple `ElementTransition`s. This field indicates the delay in starting animations for this transition on this universal timeline.
+
+* `newElement` : A reference to the [Element](https://dom.spec.whatwg.org/#interface-element) in the new Document.
+
+* `previousElement` : A reference to a `PreviousDocumentElement` provided in the `handleTransition` event (clarified below).
+
+If exactly one of `newElement` or `previousElement` is set, this is an enter or exit transition respectively. The exact animation itself is defined based on `transitionType`. If both `newElement` and `previousElement` are set, this is a paired transition with explicit start and end state. The details for element snapshots and how they are interpolated during the transition will be captured in a subsequent doc. TODO : Link when the doc is ready.
+
+### Transition Lifecycle
```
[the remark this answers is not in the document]
khushalsagar: Done.

feature_details.md
khushalsagar: Done.

feature_details.md
khushalsagar: Done.

feature_details.md
khushalsagar: Done.

feature_details.md
khushalsagar: Done.

feature_details.md
khushalsagar: Done.

feature_details.md
khushalsagar: Done.

feature_details.md
khushalsagar: Done.

feature_details.md
khushalsagar: Done.

feature_details.md
khushalsagar: Done.

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+
+Special handling is needed for a list of elements : `[CacheElement_Parent, CacheElement_Child]` where `CacheElement_Child` is a descendant of `CacheElement_Parent`. The parent element's snapshot excludes content for any descendant elements. The space covered by the descendant element shows the background which was earlier occluded by the descendant element.
+
+*TODO: Clarify the syntax for nextURL. This should allow for [URL patterns](https://github.com/WICG/urlpattern).*
+
+### New Document
+#### Referencing Elements
+The first concept for this feature on the new Document is a class : `PreviousDocumentElement`. This class provides a placeholder used by script in the new Document to reference elements from the previous Document. A `PreviousDocumentElement` object in the new Document maps 1:1 to a `CacheElement` object in the previous Document. This class has a `getPropertyMap()` API to retreive the `record<DOMString, any>` for the corresponding `CacheElement`.
+
+#### Animation Configuration
```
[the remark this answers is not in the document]
khushalsagar: Done.

feature_details.md
khushalsagar: Done.

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+    ...
+});
+```
+
+Special handling is needed for a list of elements : `[CacheElement_Parent, CacheElement_Child]` where `CacheElement_Child` is a descendant of `CacheElement_Parent`. The parent element's snapshot excludes content for any descendant elements. The space covered by the descendant element shows the background which was earlier occluded by the descendant element.
+
+*TODO: Clarify the syntax for nextURL. This should allow for [URL patterns](https://github.com/WICG/urlpattern).*
+
+### New Document
+#### Referencing Elements
```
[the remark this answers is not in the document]
khushalsagar: Done.

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+The syntax for `documentTransition.setSameOriginTransitionElements` is :
+
+```
+document.documentTransition.setSameOriginTransitionElements({
+    // nextURL: List of elements to cache.
+    USVString : [CacheElement, CacheElement],
+    USVString : [CacheElement, CacheElement],
+    ...
+});
+```
```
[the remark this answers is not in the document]
khushalsagar: Done.

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+    }
+});
+```
+Transition handling in the new Document remains the same.
+
+## API Proposal
+The proposal below is roughly divided into 3 parts : the APIs used on the old and new Document for a same origin cross-document navigation and the API for a transition within the same Document.
+
+### Previous Document
+The first part of the API is a dictionary : `CacheElement`. This is used to specify elements to cache from the previous Document. This dictionary has the following parameters :
```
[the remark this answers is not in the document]
khushalsagar: Done.

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+          });
+    }
+});
+```
+Transition handling in the new Document remains the same.
+
+## API Proposal
+The proposal below is roughly divided into 3 parts : the APIs used on the old and new Document for a same origin cross-document navigation and the API for a transition within the same Document.
+
+### Previous Document
```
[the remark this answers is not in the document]
khushalsagar: Done.

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+      document.documentTransition.setSameOriginTransitionElements({
+            e.destination.url : [{element : document.html}]
+          });
+    }
+});
+```
+Transition handling in the new Document remains the same.
+
+## API Proposal
+The proposal below is roughly divided into 3 parts : the APIs used on the old and new Document for a same origin cross-document navigation and the API for a transition within the same Document.
```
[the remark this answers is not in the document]
khushalsagar: Done.

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+    if (isSameOrigin(e.destination)) {
+      document.documentTransition.setSameOriginTransitionElements({
+            e.destination.url : [{element : document.html}]
+          });
+    }
+});
+```
+Transition handling in the new Document remains the same.
+
+## API Proposal
```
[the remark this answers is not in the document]
khushalsagar: Done.

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+
+```
+appHistory.addEventListener("navigate", e => {
+    if (isSameOrigin(e.destination)) {
+      document.documentTransition.setSameOriginTransitionElements({
+            e.destination.url : [{element : document.html}]
+          });
+    }
+});
+```
```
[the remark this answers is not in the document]
khushalsagar: Done.

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+    // image.
+    let imageAnimation = 
+      {previousElement : previousElements[1],
+       transitionType : type};
+    return [imageAnimation];
+}
+```
+
+### Integration with App History API
+This feature integrates well with the [app-history API](https://github.com/WICG/app-history) which provides developers with a central framework for all navigation related logic. The following code-snippet provides an example :
```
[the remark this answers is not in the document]
khushalsagar: Done.

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+   “foo.com/one.html”   : [{element : document.getElementById("dog"),
+                            propertyMap : {"type" : "previous"}],
+   “foo.com/three.html” : [{element : document.getElementById("dog"),
+                            propertyMap : {"type" : "next"}],
+   “foo.com/fifty.html” : [{element : document.getElementById("dog"),
+                            propertyMap : {"type" : "random"}]
+  });
+```
+
+The new Document decides the transition type based on this information :
```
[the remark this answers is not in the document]
khushalsagar: Done.

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+document.documentTransition.setSameOriginTransitionElements(
+  {
+   “foo.com/one.html”   : [{element : document.getElementById("dog"),
+                            propertyMap : {"type" : "previous"}],
+   “foo.com/three.html” : [{element : document.getElementById("dog"),
+                            propertyMap : {"type" : "next"}],
+   “foo.com/fifty.html” : [{element : document.getElementById("dog"),
+                            propertyMap : {"type" : "random"}]
+  });
+```
```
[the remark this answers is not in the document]
khushalsagar: Done.

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+    return [oldRootAnimation, newRootAnimation];
+}
+```
+
+### Old Document State
+Customizing a transition may depend on state from the old Document. The feature enables this by letting the web author pass contextual information between Documents. The following example considers a case where we use the source of navigation to decide the transition :
+
+[![Video Link for Old Dom State](https://img.youtube.com/vi/6eGmjn0o_3Q/0.jpg)](https://www.youtube.com/watch?v=6eGmjn0o_3Q)
+
+The old Document passes info about which button click initiated the navigation :
```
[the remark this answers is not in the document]
khushalsagar: Done.

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+       return [oldRootAnimation, newRootAnimation, imageAnimation];
+    }
+    
+    // Otherwise fallback to a simpler root transition.
+    return [oldRootAnimation, newRootAnimation];
+}
+```
+
+### Old Document State
+Customizing a transition may depend on state from the old Document. The feature enables this by letting the web author pass contextual information between Documents. The following example considers a case where we use the source of navigation to decide the transition :
```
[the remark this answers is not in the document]
khushalsagar: Done.

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+       
+       return [oldRootAnimation, newRootAnimation, imageAnimation];
+    }
+    
+    // Otherwise fallback to a simpler root transition.
+    return [oldRootAnimation, newRootAnimation];
+}
+```
+
+### Old Document State
```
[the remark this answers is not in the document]
khushalsagar: Done.

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+        {previousElement : previousElements[1],
+         newElement : elem};
+       
+       return [oldRootAnimation, newRootAnimation, imageAnimation];
+    }
+    
+    // Otherwise fallback to a simpler root transition.
+    return [oldRootAnimation, newRootAnimation];
+}
+```
```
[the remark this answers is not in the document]
khushalsagar: Done.

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+
+### New Document Layout
+A transition may need to be modified based on the layout of the new document for a given viewport size. Let's pick up the paired transition example again with a smaller browser window, which doesn't look as nice when the new paired element is partially onscreen :
```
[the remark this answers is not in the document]
khushalsagar: Done.

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+       
+       return [oldRootAnimation, newRootAnimation, imageAnimation];
+    }
+    
+    // Otherwise fallback to a simpler root transition.
+    return [oldRootAnimation, newRootAnimation];
+}
+```
+
+### New Document Layout
```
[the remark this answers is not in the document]
khushalsagar: Done.

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+    let imageAnimation = 
+      {previousElement : previousElements[1],
+       newElement : document.getElementById("dog")};
+       
+    return [oldRootAnimation, newRootAnimation, imageAnimation];
+}
+```
+
+### Network Latency
+Performing a seamless transition requires deferring first paint until the new Document is ready for display. Web authors should ensure that this delay is reasonable and gracefully handle slow network activity. A potential way to do this could be to update the transition based on whether a resource could be fetched within a deadline. Let's take the example of [Paired Element Transition](#paired-element-transition) again :
```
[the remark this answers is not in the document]
khushalsagar: Done.

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+    
+    let imageAnimation = 
+      {previousElement : previousElements[1],
+       newElement : document.getElementById("dog")};
+       
+    return [oldRootAnimation, newRootAnimation, imageAnimation];
+}
+```
+
+### Network Latency
```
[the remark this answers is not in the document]
khushalsagar: Done.

khushalsagar (commented):
I had to create a new PR for this review. Addressed all your comments there : https://github.com/WICG/shared-element-transitions/pull/48/files. We can continue the rest of the review there?

feature_details.md
```diff
@@ -0,0 +1,413 @@
...
+
+## Problem Statement
+When a user navigates on the web, they get to see the inner workings of web experiences: flash of white followed by a piece-meal rendering phase. This sequenced user experience results in a higher cognitive load because the user has to connect the dots between where they were, and where they are. Not only that, a smooth animation to transition between scenes also reduces the loading latency perceived by the user even if the actual loading time is the same. For these reasons, most platforms provide easy to use primitives to enable developers to build seamless transitions : [Android](https://developer.android.com/training/transitions/start-activity), [iOS/Mac](https://developer.apple.com/documentation/uikit/uimodaltransitionstyle) and [Windows](https://docs.microsoft.com/en-us/windows/apps/design/motion/page-transitions).
+
+This feature provides developers with the same capability on the web, irrespective of whether the scenes in a transition are rendered across Documents. The feature supports creating transitions between Documents which share the same [Browsing Context](https://developer.mozilla.org/en-US/docs/Glossary/Browsing_context).
```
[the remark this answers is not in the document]
jakearchibald: Yeah, it's possible to change browsing context during a regular top-level navigation, thanks to cross-origin isolation. https://web.dev/coop-coep/

I'm working on a large spec change to address a lot of this https://github.com/GoogleChrome/developer.chrome.com/pull/1175/. In this world, I'd call it the "top level traversable".

PR #58, opened 2021-09-09
Add config options documentation.

khushalsagar:


README.md
```diff
@@ -152,6 +155,33 @@ achieve the effect.
...
+    changeBodyBackground();
+    document.documentTransition.start({ sharedElements: [newE1, newE2, newE3] }).then(
+      () => console.log("transition finished"));
+  });
+}
+...
+</script>
+```
+
+* All time values are interpreted in milliseconds and must be between 0 and 5s (inclusive).
```
vmpstr, line 181:
```suggestion
* All time values are interpreted in milliseconds and must be between 0 and 5000ms (inclusive).
```

README.md
```diff
@@ -152,6 +155,33 @@ achieve the effect.
...
+* The delay is from the first frame after the start of the transition.
```
vmpstr, line 182:
```suggestion
* The delay starts from the first frame after the start of the transition.
```

README.md
```diff
@@ -152,6 +155,33 @@ achieve the effect.
...
+* The length of |sharedElementsConfig| should be the same as |sharedElements|.
```
vmpstr, line 183:
```suggestion
* The length of |sharedElementsConfig| must be the same as |sharedElements|.
```

vmpstr (commented):
Can you add a blurb about our intent of changing the API to merge the sequences. We keep it this way for now to avoid breaking existing usages

PR #63, opened 2021-10-29
Create explainer.md for rendering model

khushalsagar:


explainer.md
```diff
@@ -0,0 +1,259 @@
...
+This section describes the high level design for this feature with a tentative API:
+
+## Outgoing Document
+The transition sequence starts with the prepare phase when the browser receives a trigger for a transition. Consider the following snippet of code:
+
+```js
+<div shared-id="shared-elem"></div>
+
+// MPA version.
+addEventListener("navigate", (event) => {
```
chrishtr, line 57: Does setData cause a transition to occur? This API example is hard to understand.

explainer.md
```diff
@@ -0,0 +1,259 @@
...
+# Use-Cases
+A visual demo of the transition patterns targeted by this feature are [here](https://material.io/design/motion/the-motion-system.html#transition-patterns). The following is a summary of the semantics of these transition patterns :
+
+* Root Transitions : The full page content animates between 2 web pages with an optional static UI element on top. [Shared axis](https://material.io/design/motion/the-motion-system.html#shared-axis) shows an example.
+* Shared Element Transitions : A persistent UI element morphs into another (which could be a UI element on the next page or the whole page) changing its shape and content. [Container transform](https://material.io/design/motion/the-motion-system.html#container-transform) shows an example.
+* Entry/Exit Transitions : A UI element animates as it exits or enters the screen. This [issue](https://github.com/WICG/shared-element-transitions/issues/37) shows an example.
+
+These transitions should be feasible in SPAs (Single Page Apps) and MPAs (Multi Page Apps).
+
+# Glossary
```
chrishtr, line 15: Rather than a long list of glossary and contents at the top, let's go straight to an example. I propose you give an example of navigating from one page that is a solid color with a single rectangle shared element on it, to a page that is some other solid color with the same shared element at a different position.

Old page:
<html style="background: blue">
  <div sharedId="header" style="width: 50px; height: 50px; background: green; position: relative"></div>
</html>


New page:
  <div sharedId="header" style="width: 50px; height: 50px; background: green; position: relative; top: 300px"></div>
</html>

The API would be something like:

Old document:
addEventListener("navigate", (event) => {
  document.offerTransition({sharedElements: ["header"]});
}

New one:
<html transition="fadeIn" transitionDuration=0.1s>
...


Then you can proceed directly to the steps:

1. When offerTransition is called:
1a. Save the output of element(sharedId) for each shared id offered.
1b. Save the output of element(html).
2. Navigate to the new page, while leaving the last rendered pixels of the old page on the screen.
3. If the new page has a "transition" attribute on the HTML element, suppress rendering until DOMContentLoaded is completed.
4. After DOMContentLoaded has fired, start an animation as follows:
4a. Create a container element named and its child in the top layer for each shared element |name| in the current document. The children are accessible via a pseudo-element selector ::new-shared-element(|name|)
4b. Create a container element and child for each shared element in the old document, and identify via ::old-shared-element(|name|)
4b. Apply a UA style sheet with ... in it. (element() function, opacity:0 on shared elements in the page, etc
4c. Render as normal
5. At the completion of the animation time specified in the transitionDuration attribute, remove all pseudo-elements from the top layer.

Then say that the following changes to the element() spec would be made:

1. containment
2. no breaking
3. no visual effects
4. element(thing in old document) refers to a snapshot from 1a/1b

If you say all this in the language of the CSSWG specs, you don't need a lot of the explanatory text about what "painted content" means.

explainer.md
```diff
@@ -0,0 +1,259 @@
...
+```
+HTML (root)
+├───body
+    ├───div "shared-id=A"
+    ├───div "shared-id=C"
+```
+
+* After ready for first render, the top layer is populated with the Transition DOM for the incoming Document. The properties on the shared-container flip to the live [computed properties](#computed-properties) of the shared element in the incoming Document.
+```
+top-layer
```
chrishtr, line 146: You could use this kind of diagram, or maybe just HTML?

<top-layer>
  <shared-new>
 ...

explainer.md
```diff
@@ -0,0 +1,259 @@
...
+  box-shadow: 0px 0px 10px;
+}
+
+::shared-old(#shared-id) {
+  top: -10px;
+  left: -10px;
+}
+```
+
+An alternate approach to this is to support this natively in the browser by introducing a new content-element() function. This function would behave similar to the element() function except skipping the following properties when painting the element: box decorations and visual effects which generate a stacking context. The image will also be sized to the element's content-box (as opposed to the border-box used by the element() function). The motivation for supporting this natively would be to ensure ease of use for developers.
```
chrishtr, line 248: You should clearly separate any alternatives into an "alternatives considered" section.

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+A visual demo of the transition patterns targeted by this feature are [here](https://material.io/design/motion/the-motion-system.html#transition-patterns). The following is a summary of the semantics of these transition patterns :
+
+* Root Transitions : The full page content animates between 2 web pages with an optional static UI element on top. [Shared axis](https://material.io/design/motion/the-motion-system.html#shared-axis) shows an example.
+* Shared Element Transitions : A persistent UI element morphs into another (which could be a UI element on the next page or the whole page) changing its shape and content. [Container transform](https://material.io/design/motion/the-motion-system.html#container-transform) shows an example.
+* Entry/Exit Transitions : A UI element animates as it exits or enters the screen. This [issue](https://github.com/WICG/shared-element-transitions/issues/37) shows an example.
+
+These transitions should be feasible in SPAs (Single Page Apps) and MPAs (Multi Page Apps).
+
+# Design
+Let's take the example below which shows how the API can be used by a developer to animate the background and a shared element on a same origin navigation (MPA).
```
chrishtr, line 16: "The SPA equivalent of this case is one where the old document is mutated into the new document via DOM APIs."

explainer.md
khushalsagar: Hopefully the new example makes this clearer.

explainer.md
khushalsagar: HTML is good enough. Done.

explainer.md
khushalsagar: Done.

explainer.md
khushalsagar: One more thing, the no visual effects modification seems difficult to explain in a way that would be backwards compatible. And it also opens the discussion for why only those styles and not box decorations. I feel like if we're suggesting that, we might as well be discussing the "Natively Supported Live Animatable Properties" alternative.

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+</head>
+<body>
+  <div class="animated" shared-id="shared-header" id="header"></div>
+</body>
+</html>
+```
+
+The steps taken by the browser during the transition are as follows.
+
+1. When a navigation is initiated on the old Document, create the following pseudo elements in the top layer :
```
chrishtr, line 86:
```suggestion
1. When a navigation is initiated from the old Document to the new Document, create the following pseudo elements in the top layer:
```

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+    a. A container and child replaced element for each element with shared-id attribute. These are identified via ::shared-container(#shared-id) and ::shared-old(#shared-id) respectively.
```
chrishtr, line 87:
```suggestion
    a. A container and child replaced element for each element with the shared-id attribute. These are identified via ::shared-container(#shared-id) and ::shared-old(#shared-id) respectively.
```

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+    b. A replaced element for the root/html element identified via ::shared-old(#root).
+
+2. Apply the following UA stylesheet to the pseudo elements on the old page :
+```
+::shared-old(#root), ::shared-container(#shared-header) {
```
chrishtr, line 92: width height 100%?

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+  position: fixed;
+  top: 0px;
+  left: 0px;
+}
+
+::shared-container(#shared-header) {
+  // Container sized to the element's border-box size.
```
chrishtr, line 99:
```suggestion
  // This size is is chosen exactly according to the #shared-header element's border box dimensions after layout.
```

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+  <div class="animated" shared-id="shared-header" id="header"></div>
+</body>
+</html>
+```
+
+The steps taken by the browser during the transition are as follows.
+
+1. When a navigation is initiated on the old Document, create the following pseudo elements in the top layer :
+    a. A container and child replaced element for each element with shared-id attribute. These are identified via ::shared-container(#shared-id) and ::shared-old(#shared-id) respectively.
+    b. A replaced element for the root/html element identified via ::shared-old(#root).
```
chrishtr, line 88: 1.c The box hierarchy in the top layer is...

top layer stacking context:
  root
  container
    -> old

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+  content: element(#header);
+}
+
+::shared-old(#root) {
+  width: 100vw;
+  height: 100vh;
+  // The output of element() function on the root element.
+  content: element(html);
+}
+```
```
chrishtr, line 120: #shared-header {
  opacity: 0
}

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+
+3. Save the output of the element() function for each pseudo element referenced above with the computed size and transform applied to container elements. Then navigate to the new page leaving the last rendered pixels of the old page on screen.
```
chrishtr, line 122: Split second sentence into another step.

explainer.md
chrishtr: Alternative "element() doesn't recurse into shared element descendants"

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+
+4. When the new page loads, suppress rendering until resources required for first render have been fetched. This state is currently driven by browser heuristics but the eventual goal is to give deterministic control to developers using a [renderblocking](https://github.com/whatwg/html/issues/7131) attribute.
```
chrishtr, line 124: Remove second sentence.

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+
+5. Once the page is ready for first render, create the following pseudo elements in the top layer. This is done until the end of rAF callbacks on the page's first rendering lifecycle update[^1] :
```
chrishtr, line 126: I propose that we the top layer pseudo elements are "kept in sync with the corresponding shared elements in the DOM until the shared element transition animation completes in step 7."

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+    a. A container and child replaced element for each shared element on the old page using state saved in step 3.
+    b. A container and child replaced element for each element with shared-id attribute on the new page. Only the child element is created if a container with the matching id was already created in step a. above.
```
chrishtr, line 128: Rephrase second sentence to something like "the container is reused if already present"

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+::shared-new(#shared-header) {
+  content: element(#header);
+}
+
+// Update the container's size and transform to the shared element on the new page.
+::shared-container(#shared-header) {
+  width: 200px;
+  height: 200px;
+  transform: translate(0px, 100px);
+}
```
chrishtr, line 172: Add an animation here.

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+```
+
+7. Once no pseudo element has an active animation, remove them from the top layer.
```
chrishtr, line 175: "The animation is complete when the animation specified in step 6 completes"

explainer.md
chrishtr: ...0.5s

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+
```
chrishtr, line 176: Add a sub-section with a demo for this exactly example based on firefox stuff.

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+<html>
+<head>
+  <style>
+    body {
+      background-color: blue;
+    }
+    .animated {
+      width: 100px;
+      height: 100px;
+      background-color: red;
```
chrishtr, line 29: Put position: relative; top: 300px so it's nonzero and different than below.

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+  top: 0px;
+  left: 0px;
+}
+
+::shared-container(#shared-header) {
+  // Container sized to the element's border-box size.
+  width: 100px;
+  height: 100px;
+  
+  // A transform mapping the element to its quad in viewport space.
```
chrishtr, line 103: With my suggestion above, this should be:

transform: translate(0px, 308px)

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+  width: 200px;
+  height: 200px;
+  transform: translate(0px, 100px);
+}
+```
+
+7. Once no pseudo element has an active animation, remove them from the top layer.
+
+## Live Animatable Properties
+A common capability desirable during transitions is to interpolate styles like border-radius that change the element's shape. The [container transform](https://material.io/design/motion/the-motion-system.html#container-transform) examples show a visual demo of that. Painting properties like the element's border within its image when using the element() function makes this difficult. The following example[^2] shows how developers can control which styles are captured in the element() function vs being applied to the pseudo container on the old page.
```
chrishtr, line 178: Makes it difficult, or results in low-quality rendering? If the latter, show an example jsbin.

Nit: line wrap at < 80 chars.

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+::shared-container(#shared-header) {
+  width: 200px;
+  height: 200px;
+  transform: translate(0px, 100px);
+}
+```
+
+7. Once no pseudo element has an active animation, remove them from the top layer.
+
+## Live Animatable Properties
```
chrishtr, line 177:
```suggestion
##  Animating Box Decoration CSS Properties
```

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+This capability requires the following changes to the steps mentioned in the design above :
+* A new pseudo class ("transition") is introduced which is activated when an element is being displayed using pseudo elements in step 1 and 5.
+* When saving state in step 3, the container element's complete style is preserved instead of only the computed size and transform.
+
+## Modifications to element()
+The following changes will be made to the element() spec as a part of this proposal :
+
+* The target element must have paint containment (contain:paint) to ensure the element is the containing block for all positioned descendents and generates a stacking context.
+* The target element must disallow fragmentation (break-inside:avoid).
+* A new cached-element() function is introduced to refer to the saved output of the element() function in step 3) of the design.
```
chrishtr, line 226: Also: "nested shared elements are omitted from the output of element()"

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+    }
+  </style>
+</head>
+<body>
+  <div class="animated" shared-id="shared-header">hello world</div>
+</body>
+</html>
+```
+
+This capability requires the following changes to the steps mentioned in the design above :
```
chrishtr, line 217: You should spell out exactly what happens, this paragraph is too vague.

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+# API Examples
+## SPA
+TODO: Add example
+
+## MPA Script Based
+TODO: Add example
+
+# Alternatives Considered
+## Heirarchical Properties
+This proposal disallows a shared element to be nested inside another shared element. The restriction avoids the need to preserve the hierarchy of the shared elements and associated properties (transform, clip, effects inherited by descendents) when creating pseudo elements. This is a consideration for future iterations of the feature.
```
chrishtr, line 237: It does not allow nesting? I thought we were allowing that? It's the pseudos that don't nest in the top layer, not the shared elements right?

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+
+## Container/Child Split
+One consideration was to render each shared element using a replaced element directly instead of creating a container element. The motivation behind this split is to provide a stacking context to cross-fade the content of old and new shared elements. This was necessary to ensure blending identical pixels is a no-op using [plus-lighter](https://drafts.fxtf.org/compositing/#porterduffcompositingoperators_plus_lighter) blending. While same-origin transitions could avoid this, it enables future extensibility for cross-origin transitions where cross-fading identical images would be common.
```
chrishtr, line 240: What do you mean by "avoid this"? Do you mean "work around this"?

explainer.md
chrishtr: Also, I think you are proposing that the developer can do it themselves, with the aid of the pseudo-class. Is that correct?

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+
+## Natively Supported Live Animatable Properties
```
chrishtr, line 242:
```suggestion
## Natively Supporting Animating Box Decoration CSS Properties
```

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+An alternate approach to the setup described in [Live Animatable Properties](#live-animatable-properties) is to support this natively in the browser by introducing a new content-element() function. This function would behave similar to the element() function except skipping the following properties when painting the element: box decorations and visual effects which generate a stacking context. The image will also be sized to the element's content-box (as opposed to the border-box used by the element() function). The motivation for supporting this natively would be to make these properties animatable instead of requiring developer side changes.
```
chrishtr, line 243:
```suggestion
An alternate approach to the setup described in [Animating Box Decoration CSS Properties](#live-animatable-properties) is to support this natively in the browser by introducing a new content-element() function. This function would behave similar to the element() function except skipping the following properties when painting the element: box decorations and visual effects which generate a stacking context. The image will also be sized to the element's content-box (as opposed to the border-box used by the element() function). The motivation for supporting this natively would be to make these properties animatable instead of requiring developer side changes.
```

explainer.md
khushalsagar: I like it! This also gives the page the ability to start animations using the old content sooner (if they already know the size/location for the new elements for instance) while the new content is streaming in.

explainer.md
khushalsagar: Done.

explainer.md
vmpstr, line 87: I'd verify the formatting, since in the rendered view these lines appear in the same line as above, which makes it look very confusing.

Also, do we need to explicitly call out that these pseudo would not have an associated ::backdrop (from https://fullscreen.spec.whatwg.org/#css-pe-backdrop), I'm guessing it's fine since pseudo elements can't have pseudo elements AFAIK.

explainer.md
vmpstr, line 88: a replaced pseudo-element or an element?

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+::shared-old(#shared-header) {
+  // This is the saved output referenced in step 3.
+  content: cached-element(#header);
+}
+::shared-new(#shared-header) {
+  content: element(#header);
+}
+
+// Update the container's size and transform to the shared element on the new page.
+::shared-container(#shared-header) {
```
khushalsagar, line 168: I added the bits for isolation to get a stacking context and mix-blend-mode.

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+```
+
+3. Save the output of the element() function for each pseudo element referenced above with the computed size and transform applied to container elements. Then navigate to the new page leaving the last rendered pixels of the old page on screen.
+
+4. When the new page loads, suppress rendering until resources required for first render have been fetched. This state is currently driven by browser heuristics but the eventual goal is to give deterministic control to developers using a [renderblocking](https://github.com/whatwg/html/issues/7131) attribute.
+
+5. Once the page is ready for first render, create the following pseudo elements in the top layer. This is done until the end of rAF callbacks on the page's first rendering lifecycle update[^1] :
+    a. A container and child replaced element for each shared element on the old page using state saved in step 3.
+    b. A container and child replaced element for each element with shared-id attribute on the new page. Only the child element is created if a container with the matching id was already created in step a. above.
+    c. A replaced element for the root/html element identified via ::shared-old(#root).
```
vmpstr, line 129: Is root a reserved string in this case? Is that worth calling out? I'm not sure what's the intended behavior of 
`<div shared-id=root></div>`

explainer.md
vmpstr, line 168: You might want to move this block above `::shared-old(#shared-header)` since percent width/height depend on this width/height so it might read easier if you have this first. Up to you

explainer.md
vmpstr, line 175: Does this preclude the effect of "element doesn't move while the page underneath it shifts". Or is this an all or nothing?

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+<head>
+  <style>
+    .animated {
+      width:100px;
+      height:100px;
+      border: 10px solid black;
+      border-radius: 10% 10%;
+      box-shadow: 0px 0px 10px;
+      
+      &:transition {
```
vmpstr, line 191: Is this SASS? I think this should be pure CSS

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+  </style>
+</head>
+<body>
+  <div class="animated" shared-id="shared-header">hello world</div>
+</body>
+</html>
+```
+
+This capability requires the following changes to the steps mentioned in the design above :
+* A new pseudo class ("transition") is introduced which is activated when an element is being displayed using pseudo elements in step 1 and 5.
```
vmpstr, line 218: By activated you mean it applies when the element is being displayed using pseudo elements and doesn't apply otherwise?

Also, I'll be honest I had to keep looking back at what steps 1 and 5 and 3 (below) are, it might be nicer to reference those steps but also briefly explain what they are again

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+* When saving state in step 3, the container element's complete style is preserved instead of only the computed size and transform.
+
+## Modifications to element()
+The following changes will be made to the element() spec as a part of this proposal :
+
+* The target element must have paint containment (contain:paint) to ensure the element is the containing block for all positioned descendents and generates a stacking context.
```
vmpstr, line 224: I don't think you mentioned what a "target element" is

explainer.md
khushalsagar: Done. At 0.25s since that's the default we have right now.

explainer.md
khushalsagar: That makes it sound like we're saying the transition finishes when the UA default animation has completed but this needs to wait for developer configured animations as well. I liked the idea of letting developers specify a time or perhaps a script API (transitionFinished) if they're doing a rAF driven animation. Let's leave this open for now. I'll file a Github issue to discuss and we can update based on the conclusion there.

explainer.md
khushalsagar: You mean a link to a jsbin like this one : https://jsbin.com/xodipakazu/edit?html,output. Except for this example.

explainer.md
khushalsagar: Done.

explainer.md
khushalsagar: Done.

explainer.md
khushalsagar: Animations like this : https://jsbin.com/xodipakazu/edit?html,output (open in Firefox) are infeasible unless the box decorations are interpolated on the pseudo elements. So at least that part is necessary.

Leaving them in the snapshot is the part as well would result in low-quality rendering IMO.

explainer.md
khushalsagar: Done.

explainer.md
khushalsagar: Done. And the special cases for the root element.

explainer.md
khushalsagar: There was a discussion about this in the context of how we'll add the heirarchical aspect incrementally later. Disallowing it for now makes it easy to avoid backward compatibility issues if we change the behaviour to retain the hierarchy in pseudo elements then.

explainer.md
khushalsagar: I was tempted to merge it with the original example but keeping it seperate helps to show the additional steps required only for this functionality. Let me know what you think.

explainer.md
khushalsagar: Yes. Done.

explainer.md
khushalsagar: Done.

explainer.md
khushalsagar: Done.

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+    }
+    .animated {
+      width: 100px;
+      height: 100px;
+      background-color: red;
+    }
+  </style>
+</head>
+<body>
+  <div class="animated" shared-id="shared-header" id="header"></div>
```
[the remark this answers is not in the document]
khushalsagar: Done.

explainer.md
khushalsagar: You're right that its probably obvious that a pseudo element won't have a pseudo element of its own but I added a footnote to mention this.

explainer.md
khushalsagar: Added a comment to clarify that root is a reserved keyword.

explainer.md
khushalsagar: I thought it was clear because the point says "create the following pseudo elements"?

explainer.md
khushalsagar: Good idea. Done.

explainer.md
khushalsagar: It's all or nothing. As long as any element has an animation, all top layer elements are alive. I'm definitely open to better alternatives here. Was planning to file an issue to discuss this and update the explainer.

explainer.md
khushalsagar: Was trying to do it without copying the entire declaration. I couldn't find a way to do that with pure CSS.

explainer.md
khushalsagar: Added a comment.

explainer.md
khushalsagar: Added a comment to reiterate what the step was doing.

explainer.md
khushalsagar: Another reason I was avoiding saying we're overriding the opacity to 0 on shared elements is because this will be visible to script which can't be done in cross-origin.

explainer.md
```diff
@@ -0,0 +1,447 @@
...
+* Entry/Exit Transitions : A UI element animates as it exits or enters the screen. This [issue](https://github.com/WICG/shared-element-transitions/issues/37) shows an example.
+
+These transitions should be feasible in SPAs (Single Page Apps) and MPAs (Multi Page Apps).
+
+# Design
+Let's take the example below which shows how the API can be used by a developer to animate the background and a shared element on a same origin navigation (MPA). The SPA equivalent of this case is one where the old document is mutated into the new document via DOM APIs. See [API Extensions](#api-extensions) for more code examples.
+
+### Old Document
+```
+<html>
```
chrishtr, line 20: Add this to the document:

<a href="new-doc-url">Click me</a>

explainer.md
chrishtr: +1 to 'replaced pseudo-element' to be extra clear.

explainer.md
```diff
@@ -0,0 +1,447 @@
...
+    ```
+    ├───shared-old(root)
+    ├───shared-container(#header)
+         ├───shared-old(#header)
+    ```
+
+2. Apply the following UA stylesheet to the pseudo elements on the old page :
+```
+// "root" is a reserved keyword for the html element.
+::shared-old(#root), ::shared-container(#header-id) {
```
chrishtr, line 87: We won't be able to get away with reserved keywords here. I suggest renaming to ::shared-old-root.

explainer.md
```diff
@@ -0,0 +1,447 @@
...
+
+::shared-container(#header-id) {
+  // This size is is chosen exactly according to the #header-id element's
+  // border box dimensions after layout.
+  width: 100px;
+  height: 100px;
+  
+  // A transform positioning the element relative to the viewport so that it overlaps
+  // exactly with its screen coordinates and orientation it had on the old page.
+  transform: translate(0px, 308px);
```
chrishtr, line 108:
```suggestion
  transform: translate(8px, 308px);
```

explainer.md
chrishtr: yeah

explainer.md
```diff
@@ -0,0 +1,447 @@
...
+## Modifications to element()
+The following changes will be made to the element() spec as a part of this proposal. The element captured by this function is the target element :
+
+* The target element must have paint containment (contain:paint) to ensure the element is the containing block for all positioned descendents and generates a stacking context.
+* The target element must disallow fragmentation (break-inside:avoid).
+* A new cached-element() function is introduced to refer to the saved output of the element() function in step 3) of the design.
+* Nested shared elements are omitted from the output of element() function.
+* The special cases when running the element() function on the html element are :
+    * The natural size for the generated image is the visual viewport bounds.
+    * When creating the image, the element is drawn on a canvas with the background color of the document.
```
chrishtr, line 362: Also "shared elements captured via element() are not painted in the regular dom"

explainer.md
```diff
@@ -0,0 +1,447 @@
...
+
+* Root Transitions : The full page content animates between 2 web pages with an optional static UI element on top. [Shared axis](https://material.io/design/motion/the-motion-system.html#shared-axis) shows an example.
+* Shared Element Transitions : A persistent UI element morphs into another (which could be a UI element on the next page or the whole page) changing its shape and content. [Container transform](https://material.io/design/motion/the-motion-system.html#container-transform) shows an example.
+* Entry/Exit Transitions : A UI element animates as it exits or enters the screen. This [issue](https://github.com/WICG/shared-element-transitions/issues/37) shows an example.
+
+These transitions should be feasible in SPAs (Single Page Apps) and MPAs (Multi Page Apps).
+
+# Design
+Let's take the example below which shows how the API can be used by a developer to animate the background and a shared element on a same origin navigation (MPA). The SPA equivalent of this case is one where the old document is mutated into the new document via DOM APIs. See [API Extensions](#api-extensions) for more code examples.
```
chrishtr, line 16: Add a note that this is a possible purely declarative API shape, but is mostly to be considered a placeholder in order to describe the steps?

explainer.md
khushalsagar: This is taking me to a 404 error page. Wrong link?

explainer.md
khushalsagar: Ok. I switched it to "container pseudo-element and child replaced pseudo-element" to also make it clearer that only the child is a replaced element.

explainer.md
khushalsagar: Done.

explainer.md
khushalsagar: Done.

explainer.md
khushalsagar: Done.

explainer.md
khushalsagar: Sure, I added a note saying that the API shape is tentative. Let's avoid mentioning how much declarative vs imperative it needs to be.

explainer.md
chrishtr: I'm just suggesting you add an anchor link to the old page, pointing at the new page, so that it's a complete demo.

explainer.md
chrishtr: sure, I'm just suggesting you add a link to a demo and clarify how much is difficulty of doing something at all vs high quality

explainer.md
```diff
@@ -0,0 +1,459 @@
+# Introduction
+When a user navigates on the web, they get to see the inner workings of web experiences: flash of white followed by a piece-meal rendering phase. This sequenced user experience results in a higher cognitive load because the user has to connect the dots between where they were, and where they are. Not only that, a smooth animation to transition between scenes also reduces the loading latency perceived by the user even if the actual loading time is the same. For these reasons, most platforms provide easy to use primitives to enable developers to build seamless transitions: [Android](https://developer.android.com/training/transitions/start-activity), [iOS/Mac](https://developer.apple.com/documentation/uikit/uimodaltransitionstyle) and [Windows](https://docs.microsoft.com/en-us/windows/apps/design/motion/page-transitions).
+
+This feature provides developers with the same capability on the web, irrespective of whether the scenes in a transition are rendered across Documents.
```
chrishtr, line 4:
```suggestion
Shared Element Transitions provides developers with the same capability on the web, irrespective of whether the scenes in a transition are rendered across Documents.
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+
+# Use-Cases
+A visual demo of the transition patterns targeted by this feature are [here](https://material.io/design/motion/the-motion-system.html#transition-patterns). The following is a summary of the semantics of these transition patterns :
```
chrishtr, line 7:
```suggestion
A visual demo of some example transition patterns targeted by this feature are [here](https://material.io/design/motion/the-motion-system.html#transition-patterns). The following is a summary of the semantics of these transition patterns :
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+
+* Root Transitions : The full page content animates between 2 web pages with an optional static UI element on top. [Shared axis](https://material.io/design/motion/the-motion-system.html#shared-axis) shows an example.
```
chrishtr, line 9:
```suggestion
* Root Transitions : The full page content animates between two web pages with an optional static UI element on top. [Shared axis](https://material.io/design/motion/the-motion-system.html#shared-axis) shows an example.
```

explainer.md
```diff
@@ -0,0 +1,255 @@
...
+* Shared Element Transitions : A persistent UI element morphs into another (which could be a UI element on the next page or the whole page) changing its shape and content. [Container transform](https://material.io/design/motion/the-motion-system.html#container-transform) shows an example.
+* Entry/Exit Transitions : A UI element animates as it exits or enters the screen. This [issue](https://github.com/WICG/shared-element-transitions/issues/37) shows an example.
+
+These transitions should be feasible in SPAs (Single Page Apps) and MPAs (Multi Page Apps).
+
+# Design
+Let's take the example below which shows how the API can be used by a developer to animate the background and a shared element on a same origin navigation (MPA).
+
+### Old Document
```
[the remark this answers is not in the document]
chrishtr: ping?

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+      position: relative;
+      top: 300px;
+      width: 100px;
+      height: 100px;
+      background-color: red;
+    }
+  </style>
+</head>
+<body>
+  <div class="animated" sharedId="header-id" id="header">Shared Element</div>
```
chrishtr, line 38:
```suggestion
  <div class="animated" sharedid="header-id" id="header">Shared Element</div>
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+      background-color: green;
+    }
+    
+    ::shared-container(#header-id) {
+      animation: ::shared-container-header-id 1s ease-in;
+    }
+  </style>
+</head>
+<body>
+  <div class="animated" sharedId="header-id" id="header">Shared Element</div>
```
chrishtr, line 66:
```suggestion
  <div class="animated" sharedid="header-id" id="header">Shared Element</div>
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+</body>
+</html>
+```
+
+The steps taken by the browser during the transition are as follows.
+
+1. When the user presses "Click Me" and a navigation is initiated on the old Document, create the following pseudo-elements in the top layer[^1]. Note that a shared element must not be nested inside another shared element :
+
+    a. A container pseudo-element and child replaced pseudo-element for each element with the sharedId attribute. These are identified via ::shared-container(#sharedId) and ::shared-old(#sharedId) respectively.
```
chrishtr, line 75:
```suggestion
    a. A container pseudo-element and child replaced pseudo-element for each element with the `sharedid` attribute. These are identified via ::shared-container(shared_id) and ::shared-old(shared_id) respectively,
     where shared_id is the value of the `sharedid` attribute.
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+
+4. Navigate to the new page leaving the last rendered pixels of the old page on screen.
+
+5. When the new page loads, suppress rendering until resources required for first render have been fetched.
+
+6. Once the page is ready for first render, create the following pseudo elements in the top layer. The pseudo elements are kept in sync with the corresponding shared elements in the DOM until the transition completes as specified in step 8 :
+
+    a. A container pseudo-element and child replaced pseudo-element for each shared element on the old page using state saved in step 3.
+
+    b. A container pseudo-element and child replaced pseudo-element for each element with sharedId attribute on the new page. The container is reused if already present.
```
chrishtr, line 131:
```suggestion
    b. A container pseudo-element and child replaced pseudo-element for each element with the `sharedid` attribute on the new page. The container is reused if already present.
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+
+@keyframes ::shared-old-fade-out {
+  0% {opacity: 1;}
+  100% {opacity: 0;}
+}
+::shared-old-root, ::shared-old(#header-id) {
+  animation: ::shared-old-fade-out 0.25s;
+}
+
+// Generated for each shared element with the syntax shared-container-sharedId.
```
chrishtr, line 208:
```suggestion
// Generated for each shared element with the syntax shared-container-shared_d.
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+    
+    // Offset the image to account for the border in the snapshot.
+    ::shared-old(#header-id) {
+      top: -10px;
+      left: -10px;
+    }
+  </style>
+</head>
+<body>
+  <div class="animated" sharedId="header-id" id="header">Shared Element</div>
```
chrishtr, line 270:
```suggestion
  <div class="animated" sharedid="header-id" id="header">Shared Element</div>
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+      }
+    }
+    ::shared-old(#header-id), ::shared-new(#header-id) {
+      animation: position-animation 1s ease-in;
+    }
+    
+  </style>
+</head>
+<body>
+  <div class="animated" sharedId="header-id" id="header">Shared Element</div>
```
chrishtr, line 341:
```suggestion
  <div class="animated" sharedid="header-id" id="header">Shared Element</div>
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+* The async callback initiates load of the next page and initiates step 4 which suppresses rendering.
+* When the callback returns, the new DOM is considered ready for first render. This starts step 5 onwards to create new pseudo elements and start animations.
+
+## Additional Script APIs
+The following example shows how developers can configure the transition in script for an MPA.
+
+### Old Document
+```js
+addEventListener("navigate", (event) => {
+  // Add sharedId attribute to elements to offer for the transition
```
chrishtr, line 400:
```suggestion
  // Add sharedid attribute to elements to offer for the transition
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+  // based on current document state.
+  document.querySelector(".header").sharedId="header-id";
```
chrishtr, line 402:
```suggestion
  document.querySelector(".header").sharedid="header-id";
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+requestAnimationFrame(() => {
+  let pendingTransition = document.documentTransition.getPendingTransition();
+  if (pendingTransition.getData().version !== 123)
+    return;
+  
+  // |offeredTransitionElements| provides a list of objects to access state
+  // saved from the old page.
+  let oldHeader = pendingTransition.offeredTransitionItems.get("header-id");
+  if (oldHeader) {
+    // Add sharedId attribute to elements animated in the new DOM.
```
chrishtr, line 421:
```suggestion
    // Add sharedid attribute to elements animated in the new DOM.
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+    document.querySelector(".header").sharedId="header-id";
```
chrishtr, line 422:
```suggestion
    document.querySelector(".header").sharedid="header-id";
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+    }
+    .animated {
+      position: relative;
+      top: 100px;
+      width: 200px;
+      height: 200px;
+      background-color: green;
+    }
+    
+    ::shared-container(#header-id) {
```
chrishtr, line 60:
```suggestion
    ::shared-container(header-id) {
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+}
+
+::shared-old-root {
+  width: 100vw;
+  height: 100vh;
+  // The output of element() function on the root element.
+  content: element(html);
+}
+
+::shared-container(#header-id) {
```
chrishtr, line 102:
```suggestion
::shared-container(header-id) {
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+  // border box dimensions after layout.
+  width: 100px;
+  height: 100px;
+  
+  // A transform positioning the element relative to the viewport so that it overlaps
+  // exactly with its screen coordinates and orientation it had on the old page.
+  transform: translate(8px, 308px);
+}
+
+::shared-old(#header-id) {
```
chrishtr, line 113:
```suggestion
::shared-old(header-id) {
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+    ├───shared-new-root
+    ├───shared-old-root
+    ├───shared-container(#header)
+         ├───shared-new(#header)
+         ├───shared-old(#header)
+    ```
+
+7. Apply the following UA stylesheet to the pseudo elements on the new page.
+```
+::shared-old-root, ::shared-new-root, ::shared-container(#header-id) {
```
chrishtr, line 147:
```suggestion
::shared-old-root, ::shared-new-root, ::shared-container(header-id) {
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+  width: 200px;
+  height: 200px;
+  transform: translate(8px, 108px);
+  
+  // The blend mode referenced below is not currently exposed with mix-blend-mode.
+  isolation: isolate;
+  mix-blend-mode: [plus-lighter](https://drafts.fxtf.org/compositing/#porterduffcompositingoperators_plus_lighter);
+}
+
+::shared-old(#header-id), ::shared-new(#header-id) {
```
chrishtr, line 170:
```suggestion
::shared-old(#header-id), ::shared-new(header-id) {
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+::shared-new(#header-id) {
+  content: element(#header);
+}
+
+// Default animations added by the UA which can be overridden by the developer stylesheet/script.
+@keyframes ::shared-new-fade-in {
+  0% {opacity: 0;}
+  100% {opacity: 1;}
+}
+::shared-new-root, ::shared-new(#header-id) {
```
chrishtr, line 196:
```suggestion
::shared-new-root, ::shared-new(header-id) {
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+  animation: ::shared-new-fade-in 0.25s;
+}
+
+@keyframes ::shared-old-fade-out {
+  0% {opacity: 1;}
+  100% {opacity: 0;}
+}
+::shared-old-root, ::shared-old(#header-id) {
```
chrishtr, line 204:
```suggestion
::shared-old-root, ::shared-old(header-id) {
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+
+// Generated for each shared element with the syntax shared-container-sharedId.
+@keyframes ::shared-container-header-id {
+  from {
+    width: 100px;
+    height: 100px;
+    transform: translate(0px, 308px);
+  }
+}
+::shared-container(#header-id) {
```
chrishtr, line 216:
```suggestion
::shared-container(header-id) {
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+      box-shadow: 0px 0px 10px;
+      &:transition {
+        /* Retain the border to ensure it is painted transparent but the box size is unchanged. */
+        border: 10px solid transparent;
+        border-radius: none;
+        box-shadow: none;
+      }
+    }
+    
+    ::shared-container(#header-id) {
```
chrishtr, line 256:
```suggestion
    ::shared-container(header-id) {
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+      border: 10px solid black;
+      border-radius: 10% 10%;
+      box-shadow: 0px 0px 10px;
+    }
+    
+    // Offset the image to account for the border in the snapshot.
+    ::shared-old(#header-id) {
```
chrishtr, line 263:
```suggestion
    ::shared-old(header-id) {
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+      box-shadow: 0px 0px 5px;
+      
+      &:transition {
+        border: 5px solid transparent;
+        border-radius: none;
+        box-shadow: none;
+      }
+    }
+    
+    ::shared-container(#header-id) {
```
chrishtr, line 302:
```suggestion
    ::shared-container(header-id) {
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+        border: 5px solid black;
+        border-radius: 50% 50%;
+        box-shadow: 0px 0px 5px;
+    }
+    
+    ::shared-new(#header-id) {
```
chrishtr, line 308:
```suggestion
    ::shared-new(header-id) {
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+    }
+    
+    @keyframes border-animation {
+      from {
+        border: 10px solid black;
+        border-radius: 10% 10%;
+        box-shadow: 0px 0px 10px;
+      }
+    }
+    ::shared-container(#header-id) {
```
chrishtr, line 320:
```suggestion
    ::shared-container(header-id) {
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+      from {
+        top: -10px;
+        left: -10px;
+      }
+      to {
+        top: -5px;
+        left: -5px;
+      }
+    }
+    ::shared-old(#header-id), ::shared-new(#header-id) {
```
chrishtr, line 334:
```suggestion
    ::shared-old(header-id), ::shared-new(header-id) {
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+  top: -10px;
+  left: -10px;
+}
+```
+
+## Modifications to element()
+The following changes will be made to the element() spec as a part of this proposal. The element captured by this function is the target element :
+
+* The target element must have paint containment (contain:paint) to ensure the element is the containing block for all positioned descendents and generates a stacking context.
+* The target element must disallow fragmentation (break-inside:avoid).
```
chrishtr, line 370:
```suggestion
* The target element must disallow fragmentation (similar to `break-inside:avoid`).
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+* A new cached-element() function is introduced to refer to the saved output of the element() function in step 3 of the design.
```
chrishtr, line 371:
```suggestion
* A new cached-element() function is introduced to refer to the saved output of the element() function in step 3 of the design. (This part may not be developer exposed though.)
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+* Nested shared elements are omitted from the output of element() function.
+* Elements captured using element() and displayed via pseudo-elements during the transition are not painted in the regular DOM.
```
chrishtr, line 373:
```suggestion
* Elements captured using element() and displayed via pseudo-elements during the transition are not painted in the regular DOM - they behave as if they have `content-visibility: hidden`, except that they don't have `contain:size`.
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+  }
+});
+```
+
+# Alternatives Considered
+## Heirarchical Properties
+This proposal disallows a shared element to be nested inside another shared element. The restriction avoids the need to preserve the hierarchy of the shared elements and associated properties (transform, clip, effects inherited by descendents) when creating pseudo elements. This is a consideration for future iterations of the feature.
+
+## Container/Child Split
+One consideration was to render each shared element using a replaced element directly instead of creating a container element. The motivation behind this split is to provide a stacking context to cross-fade the content of old and new shared elements. This was necessary to ensure blending identical pixels is a no-op using [plus-lighter](https://drafts.fxtf.org/compositing/#porterduffcompositingoperators_plus_lighter) blending. While same-origin transitions could work around this, it enables future extensibility for cross-origin transitions where cross-fading identical images would be common.
```
chrishtr, line 445:
```suggestion
One consideration is to render each shared element using a replaced element directly instead of creating a container element. The motivation behind this split is to provide a stacking context to cross-fade the content of old and new shared elements. This was necessary to ensure blending identical pixels is a no-op using [plus-lighter](https://drafts.fxtf.org/compositing/#porterduffcompositingoperators_plus_lighter) blending. While same-origin transitions could work around this, it enables future extensibility for cross-origin transitions where cross-fading identical images would be common.
```

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+
+## Natively Supporting Animating Box Decoration CSS Properties
+An alternate approach to the setup described in [Animating Box Decoration CSS Properties](#live-animatable-properties) is to support this natively in the browser by introducing a new content-element() function. This function would behave similar to the element() function except skipping the following properties when painting the element: box decorations and visual effects which generate a stacking context. The image will also be sized to the element's content-box (as opposed to the border-box used by the element() function). The motivation for supporting this natively would be to make these properties animatable instead of requiring developer side changes.
```
chrishtr, line 448:
```suggestion
An alternate approach to the setup described in [Animating Box Decoration CSS Properties](#live-animatable-properties) is to support this natively in the browser by introducing a new `content-element()` function. This function would behave similarl to the `element()` function except skipping the following properties when painting the element: box decorations and visual effects which generate a stacking context. The image will also be sized to the element's content-box (as opposed to the border-box used by the element() function). The motivation for supporting this natively would be to make these properties animatable instead of requiring developer side changes.
```

chrishtr (commented):
Please reflow to < 80 columns to make it easier to provide suggestions.

explainer.md
khushalsagar: Fixed.

explainer.md
```diff
@@ -0,0 +1,459 @@
...
+
+# Security/Privacy Considerations
+The security considerations below are limited to same-origin transitions :
```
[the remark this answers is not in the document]
khushalsagar: Clarified that they apply to cross-origin as well. But are a subset of the considerations for cross-origin.

chrishtr (approved):

explainer.md
```diff
@@ -0,0 +1,600 @@
...
+and
+[Windows](https://docs.microsoft.com/en-us/windows/apps/design/motion/page-transitions).
+
+Shared Element Transitions provides developers with the same capability on the
+web, irrespective of whether the transitions are cross-document or
+intra-document (SPA).
+
+# Use-Cases
+
+A visual demo of some example transition patterns targeted by this feature are
```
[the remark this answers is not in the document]
vmpstr: typo? :)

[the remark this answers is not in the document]
khushalsagar: Typo indeed. I'll resolve this.

PR #69, opened 2021-11-09
Update explainer.md based on Jake's review.

khushalsagar:


khushalsagar:
Github won't let me comment on unchanged lines. :/ So replying to a comments here :

"Doesn't Chrome (and possibly other browsers) avoid the flash of white in many cases with things like paint-holding?"
It's opportunistic, heuristic driven and we're very conservative about how long we hold the first paint for. So it's gotten rarer but can still happen. I don't think this is an important point though so we can remove or restructure (suggestions welcome for the latter).

explainer.md
```diff
@@ -24,9 +24,10 @@ A visual demo of some example transition patterns targeted by this feature are
 The following is a summary of the semantics of these transition patterns :
 
 *   Root Transitions : The full page content animates between two web pages with
-    an optional static UI element on top.
+    optional static UI elements on top.
     [Shared axis](https://material.io/design/motion/the-motion-system.html#shared-axis)
-    shows an example.
+    shows these examples. #1 and #2 have a transition with static elements while #3
```
khushalsagar, line 29: "I found this a little confusing at first, since what you're describing isn't 'shared axis'. Although examples 1 & 2 in the video do have a static component, example 3 doesn't."

I intentionally stated *optional* static UI elements so it was clear that shared axis patterns can be done without a static element as well.

explainer.md
```diff
@@ -606,6 +640,10 @@ subset of what's required for cross-origin transitions :
     already has knowledge of the cross-origin policy on the new page. They can
     ensure not to reference cross-origin resources in the properties made live.
 
+    Note that painting cross-origin resources in the image generated by element()
+    is not a security concern since the pixel content will never be accessible to
+    the incoming page.
```
jakearchibald, line 645: I wonder if that is a Chrome implementation thing. I mean, there must be some reason we changed the behaviour of `` on isolated pages.

jakearchibald (approved):

explainer.md
```diff
@@ -578,15 +582,44 @@ images would be common.
...
+The motivation for supporting this natively would be to make these properties
+animatable by default instead of requiring developers to implement it themselves.
+This is a consideration for future iterations of the feature.
+
+## Clipping of Effects
+
+The decorated bounding box used by the element() function clips effects which paint
+outside the element's border-box (like box-shadow, drop-shadow) when generating an
+element's image. Developers can avoid this by using the approach outlined in
+[Animating Box Decoration CSS Properties](#animating-box-decoration-css-properties) to
```
jakearchibald, line 602: That doesn't solve the problem for things like text-shadow, eg https://static-misc-3.glitch.me/overflow-demo/

jakearchibald, line 602: That doesn't solve the problem for things like text-shadow, eg https://static-misc-3.glitch.me/overflow-demo/

PR #95, opened 2021-11-24
Update explainer.md with resolved open issues

khushalsagar:


explainer.md
```diff
@@ -50,13 +50,13 @@ Aside from the root, an element offered for a transition has the following restr
...
 - [`break-inside: avoid`](https://developer.mozilla.org/en-US/docs/Web/CSS/break-inside) which disallows fragmentation ensuring the element content is a single rect, i.e., it doesn't break across lines or columns, again allowing the element to be treated as a single unit.
 
-- [Open question](https://github.com/WICG/shared-element-transitions/issues/71): If the developer tries to break the above rules, what should happen? Is paint containment applied by the browser? Is that element dropped from the transition? Is the whole transition abandoned?
+The developer must add the properties above to an element's style if its offered for a transition. If this constraint is not satisfied at any point during the transition, the transition is aborted. See [issue](https://github.com/WICG/shared-element-transitions/issues/71) for detailed discussion.
```
vmpstr, line 53: Presumably this would only not be satisfied (or detected as not satisfied) after style resolution

vmpstr: It's also unclear to me if we're requiring contain: paint or simply a property which adds paint containment. I'll file a separate issue for this.

explainer.md
```diff
@@ -89,21 +89,25 @@ The second mode where styles are copied to a container element won't be part of
 
 ### Nested transition elements
 
-In the example transition, the content of the header cross-fades from Page-A to Page-B. An even smoother transition could be achieved by also animating the site title and avatar 'chip' independently. However, 'v1' of this proposal disallows a shared element to be nested inside another shared element.
+In the example transition, the content of the header cross-fades from Page-A to Page-B. An even smoother transition could be achieved by also animating the site title and avatar 'chip' independently. When a developer offers nested shared elements, the state can be captured in the following two modes :
```
vmpstr, line 92:
```suggestion
In the example transition, the content of the header cross-fades from Page-A to Page-B. An even smoother transition could be achieved by also animating the site title and avatar 'chip' independently. When a developer offers nested shared elements, the state can be captured in one of the following two modes:
```

explainer.md
```diff
@@ -89,21 +89,25 @@ The second mode where styles are copied to a container element won't be part of
...
 
-The restriction avoids the need to preserve the hierarchy of shared elements and associated properties (transform, clip, effects inherited by descendants) in the DOM representation created to render the images referenced above. This helps in minimizing the scope for 'v1' of the feature.
+### Flat List Mode
```
vmpstr, line 94:
```suggestion
#### Flat List Mode
```

explainer.md
```diff
@@ -89,21 +89,25 @@ The second mode where styles are copied to a container element won't be part of
...
+The content of all nested shared elements is removed when generating the image for a shared element so they can be animated as independent images. This is similar to removing the content of shared elements from the root image.
 
-- [Open question](https://github.com/WICG/shared-element-transitions/issues/74): If the developer tries to do this, what should happen, and when is that verified? Is the nested offering ignored? Is the whole transition abandoned?
+In addition to the element's image, the browser also caches the element's border-box bounds and a transform positioning the element relative to the viewport so that it overlaps exactly with its quad on the old page.
+
+## Hierarchy Mode
```
vmpstr, line 99:
```suggestion
#### Hierarchy Mode
```

explainer.md
```diff
@@ -89,21 +89,25 @@ The second mode where styles are copied to a container element won't be part of
...
+In this mode, the hierarchy of shared elements is preserved along with a transform which is relative to its nearest ancestor shared element instead of the viewport. When used with the "computed style + content image" mode, this also preserves other hierarchical properties (clip, effects inherited by descendants) that a shared element inherits from its ancestor.
```
vmpstr, line 100:
```suggestion
In this mode, the hierarchy of shared elements is preserved along with a transform which is relative to its nearest ancestor shared element instead of the viewport. When used with the "computed style + content image" mode, this also preserves other hierarchical properties (clip, inherited effects) that a shared element inherits from its ancestor. 
```

explainer.md
```diff
@@ -89,21 +89,25 @@ The second mode where styles are copied to a container element won't be part of
...
 
-- [Open question](https://github.com/WICG/shared-element-transitions/issues/74): If the developer tries to do this, what should happen, and when is that verified? Is the nested offering ignored? Is the whole transition abandoned?
+In addition to the element's image, the browser also caches the element's border-box bounds and a transform positioning the element relative to the viewport so that it overlaps exactly with its quad on the old page.
+
+## Hierarchy Mode
+In this mode, the hierarchy of shared elements is preserved along with a transform which is relative to its nearest ancestor shared element instead of the viewport. When used with the "computed style + content image" mode, this also preserves other hierarchical properties (clip, effects inherited by descendants) that a shared element inherits from its ancestor. 
+
+This allows nested shared elements to continue to be positioned and clipped relative to their ancestor shared elements during a transition. This is proposed as a separate mode to minimize the scope for 'v1' of the feature. See [issue](https://github.com/WICG/shared-element-transitions/issues/74) for detailed discussion.
```
vmpstr, line 102: Is this a separate mode solely for minimizing the scope of the v1?

explainer.md
```diff
@@ -221,7 +235,8 @@ document.performTransition((transition) => {
 This sketch is particularly half-baked. A more concrete proposal will be possible when more of the concepts are decided.
 
 - [Open question](https://github.com/WICG/shared-element-transitions/issues/79): Do we need `createTransitionElement`? It could be done via adding CSS properties, but it might be clumsy if the developer is going to immediately remove the properties afterwards.
-- [Open question](https://github.com/WICG/shared-element-transitions/issues/80): How does the outgoing page offer just the root for transition?
+- [Open question](https://github.com/WICG/shared-element-transitions/issues/80): How does the outgoing page offer just the
```
ianvollick, line 238: nit: is this an unintentional edit?

explainer.md
```diff
@@ -121,10 +125,9 @@ transition element
 - **image wrapper**: This element has a width and height of 100%, and [`isolation: isolate`](https://developer.mozilla.org/en-US/docs/Web/CSS/isolation). This wrapper is useful when cross-fading images (documented later).
 - **image**: This contains the cached image, which may paint outside the parent elements. This would be a replaced element so CSS properties like `object-fit` will be supported. This element has a width and height of 100%, although the image may paint outside of its own bounds, similar to how a `box-shadow` is painted outside of an element's bounds.
 
-These elements will be addressable via pseudo-elements, although they may be exposed as full elements via a JS API.
+These elements will be addressable via pseudo-elements.
```
vmpstr, line 128: not sure what "addressable" means in this context, but this was there before

explainer.md
```diff
@@ -147,7 +150,18 @@ The root elements of each page are automatically associated.
 
 Note that the order in which the transition elements are painted can be configured by UA and/or developer stylesheets using z-index.
 
-- [Open question](https://github.com/WICG/shared-element-transitions/issues/23): How should the default UA animation order these elements? And also handle a change in associated elements between the 2 pages. 
+- [Open question](https://github.com/WICG/shared-element-transitions/issues/23): How should the default UA animation order these elements? And also handle a change in associated elements between the 2 pages.
+
+### How are transition elements painted?
+During the transition a new stacking context (called uber-root) is created with the following hierarchy :
```
vmpstr, line 156: I don't really like the name uber-root for official explainers. I don't have a better name, but maybe we can just not name it?

explainer.md
```diff
@@ -147,7 +150,18 @@ The root elements of each page are automatically associated.
...
+
+```
+uber-root stacking context
+└─ root stacking context
```
vmpstr, line 160: Perhaps root stacking context can be "page root stacking context" or something that makes it clear that it's a root for all things that the page itself may paint. Although technically the transition stacking context's elements are also styled by the page, so nevermind...

khushalsagar: It's actually better to call it "root stacking context" to make it obvious that we're talking about the existing concept of a page's root stacking context?

explainer.md
khushalsagar: I mostly did it for the diagram below to make the proposed hierarchy obvious. In general, we have a lot of naming in this explainer that needs to change. :)

explainer.md
khushalsagar: How about now? The aim is to say that developers can access them as pseudo elements.

explainer.md
khushalsagar: That's the main reason, yes. Other reason is that we can choose to not support it for the out of process cross-origin implementation. But the primary motivation is to scope it out of v1.

explainer.md
khushalsagar: Yes, we can detect it during style resolution and abort in the same frame.

explainer.md
khushalsagar: Yes, my bad. Fixed.

explainer.md
vmpstr: Do you mind saying this? "At any point during transition" sounds too vague

khushalsagar: Done.

explainer.md
ianvollick: Would it be possible to support this cross-origin if hierarchical info were sent to wherever it is that interpolation happens?

khushalsagar: Yeah, it's definitely possible. We'll need to build a data structure that preserves the transform, clip and stacking context hierarchy of these elements. All of those should be the same since with paint containment each shared element creates a stacking context anyway.

One aspect worth pointing out is that this hierarchy mode will only retain positioning and clipping from an ancestor. If the ancestor has effects like opacity (which should continue to be inherited by the descendants) then we have to ensure that the opacity is "live" and not baked into the snapshot. So it applies to the shared element's image and also to the images of its descendants.

@chrishtr has also mentioned not painting visual effects into the snapshot. This is a divergence from the element() function behaviour but maybe it makes sense to have the "As a single image" mode be "As style for visual effects + single image" mode instead. The styles for visual effects which apply to descendants are copied to the container.

We can support "As style for visual effects + single image" + hierarchy for cross-origin but supporting "As the element's computed style + content image" for cross-origin will be too difficult.

vmpstr (approved):

explainer.md
khushalsagar: I've filed a WICG issue for the question about visual effects above.

ianvollick: Thanks for the extra info. I was just hoping to clarify the point raised earlier and didn't mean to hold up the review. I will stamp and defer to Vlad's review.

ianvollick (approved):

explainer.md
```diff
@@ -50,13 +50,13 @@ Aside from the root, an element offered for a transition has the following restr
...
 
-- [Open question](https://github.com/WICG/shared-element-transitions/issues/71): If the developer tries to break the above rules, what should happen? Is paint containment applied by the browser? Is that element dropped from the transition? Is the whole transition abandoned?
+The developer must add the properties above to an element's style if its offered for a transition. This constraint is verified during style resolution for all rendering lifecycle updates during a transition. If not satisfied, the transition is aborted. See [issue](https://github.com/WICG/shared-element-transitions/issues/71) for detailed discussion.
 
 When a developer offers elements for a transition, there are two modes they can choose from:
 
 ### As a single image
 
-The entire painting of the element is captured, including things which appear outside of its bounding box such as shadows and blurs, as a single CSS image. The resulting dimensions, transform, and viewport position is also captured so the image can be correctly positioned later.
+The entire painting of the element is captured, including things which appear outside of its bounding box such as shadows and blurs, as a single CSS image.
```
jakearchibald, line 59: This change feels like an error. It means:

- Transition elements will appear at 0,0 all stacked on top of each other, since viewport position is no longer captured.
- Transition elements will have to be sized manually, since their dimensions are no longer captured.
- If the transition element is rotated 45 degs, that will be baked into the capture.

explainer.md
```diff
@@ -89,21 +89,25 @@ The second mode where styles are copied to a container element won't be part of
 
 ### Nested transition elements
 
-In the example transition, the content of the header cross-fades from Page-A to Page-B. An even smoother transition could be achieved by also animating the site title and avatar 'chip' independently. However, 'v1' of this proposal disallows a shared element to be nested inside another shared element.
+In the example transition, the content of the header cross-fades from Page-A to Page-B. An even smoother transition could be achieved by also animating the site title and avatar 'chip' independently. When a developer offers nested shared elements, the state can be captured in one of the following two modes:
 
-The restriction avoids the need to preserve the hierarchy of shared elements and associated properties (transform, clip, effects inherited by descendants) in the DOM representation created to render the images referenced above. This helps in minimizing the scope for 'v1' of the feature.
+#### Flat List Mode
+The content of all nested shared elements is removed when generating the image for a shared element so they can be animated as independent images. This is similar to removing the content of shared elements from the root image.
```
jakearchibald, line 95: I can't get my head around this. Is the word "content" being used correctly here?

```html
<div>Hello</div>
```

The content of that `div` is the `Hello` text node. If the div has a border or background colour, that isn't part of the content.

explainer.md
```diff
@@ -89,21 +89,25 @@ The second mode where styles are copied to a container element won't be part of
 
 ### Nested transition elements
 
-In the example transition, the content of the header cross-fades from Page-A to Page-B. An even smoother transition could be achieved by also animating the site title and avatar 'chip' independently. However, 'v1' of this proposal disallows a shared element to be nested inside another shared element.
+In the example transition, the content of the header cross-fades from Page-A to Page-B. An even smoother transition could be achieved by also animating the site title and avatar 'chip' independently. When a developer offers nested shared elements, the state can be captured in one of the following two modes:
 
-The restriction avoids the need to preserve the hierarchy of shared elements and associated properties (transform, clip, effects inherited by descendants) in the DOM representation created to render the images referenced above. This helps in minimizing the scope for 'v1' of the feature.
+#### Flat List Mode
```
jakearchibald, line 94: I'm not sure what this section is describing. You can try to explain it to me in a chat 😄

explainer.md
jakearchibald: It isn't clear to me what this is vs "root image"

explainer.md
```diff
@@ -279,21 +293,19 @@ CSS can be used to make changes to the default animation, or override `animation
...
-
-This isn't unusual, as developer currently use things like [`::placeholder`](https://developer.mozilla.org/en-US/docs/Web/CSS/::placeholder) to address things in a UA shadow root within `<input>`. We can't expose the shadow root within `<input>`, because the structure is non-standard, however, for transitions, the structure will be standardized, so it can be exposed.
+So far the transition elements have been addressed by pseudo-element selectors, but JavaScript could be given access to the elements using the [CSSPseudoElement](https://drafts.csswg.org/css-pseudo-4/#CSSPseudoElement-interface) interface.
 
 ```js
 // In Page-B
 document.performTransition((transition) => {
   // Build up transition parts, then…
-  transition.root.querySelector('[part=header]').animate(…);
+  document.documentElement.pseudo("page-transition-container(name)").animate(...);
```
jakearchibald, line 302: This API doesn't exist. Is it being proposed here?

explainer.md
```diff
@@ -279,21 +293,19 @@ CSS can be used to make changes to the default animation, or override `animation
...
 });
 ```
 
 - [Open question](https://github.com/WICG/shared-element-transitions/issues/86) What's the deadline for calling `performTransition`?
 
-If the 'stage' of the transition is exposed as a shadow root like this, the developer can interact with the elements in a regular way. The developer could even create elements specifically for the transition.
+Note that using pseudo-elements here implies that the developer can not change the DOM structure of transition elements. An alternate approach considered to provide this extensibility uses shadow DOM. The pseudo-element option was preferred to make it easier to build on concepts which already exist in the platform and ensure ease of implementation. See [issue](https://github.com/WICG/shared-element-transitions/issues/93) for detailed discussion.
```
jakearchibald, line 308: This feels like it's saying shadow DOM doesn't already exist in the platform.

explainer.md
jakearchibald: (we chatted and I think I understand it now. I'll add to this PR)

explainer.md
jakearchibald: (we chatted and I think I understand it now. I'll add to this PR)

explainer.md
jakearchibald: (we chatted and I think I understand it now. I'll add to this PR)

explainer.md
khushalsagar: You mean the .animate API on CSSPseudoElement? Yeah, I was proposing that.

I've noticed a bunch of APIs for pseudo elements work by adding it as an argument, like window.getComputedStyle(element, "::after"). This is likely there wasn't an object in script for a pseudo element when these were added. Now that we have one, why not add these APIs to that interface.

explainer.md
khushalsagar: The concept that's missing is creating a shadow DOM which has no element to be its shadow host.

explainer.md
hvanops: "This is proposed as a separate mode to minimize the scope for 'v1' of the feature." This makes it sound like separating into two modes minimizes the scope. Can we clarify that only Flat List mode will be targeted for v1 to minimize scope.

khushalsagar: @jakearchibald mentioned that he'll reword this section to make the details for the 2 modes clearer. Will make sure that flat list being the only mode for v1 is obvious after his edits.

explainer.md
```diff
@@ -89,21 +89,29 @@ The second mode where styles are copied to a container element won't be part of
...
+In the example transition, the content of the header cross-fades from Page-A to Page-B. An even smoother transition could be achieved by also animating the site title and avatar 'chip' independently. To allow for this, an offered element can contain other offered elements.
 
-The restriction avoids the need to preserve the hierarchy of shared elements and associated properties (transform, clip, effects inherited by descendants) in the DOM representation created to render the images referenced above. This helps in minimizing the scope for 'v1' of the feature.
+When an element is captured, its painting is 'lifted' from the parent offered element, which may be the root. This is similar to how browsers handle composited elements.
 
-- [Open question](https://github.com/WICG/shared-element-transitions/issues/74): If the developer tries to do this, what should happen, and when is that verified? Is the nested offering ignored? Is the whole transition abandoned?
+### Retaining hierarchy during a transition
```
jakearchibald, line 96: @khushalsagar I reworked this idea a bit so it feels less like a v2 switch, and more like opting into a new positioning space like `position: relative`. Let me know what you think. I went back and forth on elements of this, so I'm absolutely open to tweaks and better ideas.

explainer.md
```diff
@@ -89,21 +89,29 @@ The second mode where styles are copied to a container element won't be part of
...
-
-The top layer content ensures that the user continues to see Page-A's visuals as Page-B is loading. Note that this may not reproduce the exact rendering on Page-A. For example, the relative paint order of shared elements is preserved in the top layer. But if a shared element was occluded by another element, the latter is painted into the root's image unless it is also offered as a shared element.
+The stage is a viewport-filling container with fixed position. The offered elements from Page-A are positioned absolutely at (0,0), nested according in their closest 'transition container' or the stage if there's no parent container, and moved into their previous viewport-relative positions using the cached transform. Their content is painted on top of Page-B which ensures that the user continues to see Page-A's visuals as Page-B is loading. Note that this may not reproduce the exact rendering on Page-A. For example, the relative paint order of shared elements is preserved when rendered on top of Page-B. But if a shared element was occluded by another element, the latter is painted into the root's image unless it is also offered as a shared element.
```
jakearchibald, line 114: @khushalsagar in my last PR I had the offered elements absolutely positioned in a fixed position container, and we switched it back to everything being fixed position. However, I think we need absolute positioning else the nesting doesn't make sense. We want transition elements to be positioned relative to 0,0 of their transition container, which may be the viewport, but may not be. Making them fixed position would put everything at 0,0 in viewport space, which doesn't seem to work.

explainer.md
```diff
@@ -113,41 +121,60 @@ The CSS images and computed properties/styles cached from Page-A are represented
 
 ```
 transition element
-└─ image wrapper
-   └─ image
+└─ states
+   └─ state
```
jakearchibald, line 125: I've added a new wrapper 😬, but it seems necessary to handle nesting. I guess we could omit this wrapper for transition elements that aren't a 'transition container', but is it better to always have it for consistency?

explainer.md
ianvollick: Regarding fixed positioning; the containing block for a fixed position element is usually the viewport, but not always (eg, setting `transform:` on an element establishes a containing block for all positioned descendants, including fixed [1]). If we want to describe elements as being fixed with respect to its nearest shared ancestor, I think we could use the same type of prose.

[1] https://www.w3.org/TR/css-transforms-1/#:~:text=the%20transform%20property%20also%20causes%20the%20element%20to%20establish%20a%20containing%20block%20for%20all%20descendants.

explainer.md
```diff
@@ -89,21 +89,29 @@ The second mode where styles are copied to a container element won't be part of
 
 ### Nested transition elements
 
-In the example transition, the content of the header cross-fades from Page-A to Page-B. An even smoother transition could be achieved by also animating the site title and avatar 'chip' independently. However, 'v1' of this proposal disallows a shared element to be nested inside another shared element.
+In the example transition, the content of the header cross-fades from Page-A to Page-B. An even smoother transition could be achieved by also animating the site title and avatar 'chip' independently. To allow for this, an offered element can contain other offered elements.
 
-The restriction avoids the need to preserve the hierarchy of shared elements and associated properties (transform, clip, effects inherited by descendants) in the DOM representation created to render the images referenced above. This helps in minimizing the scope for 'v1' of the feature.
+When an element is captured, its painting is 'lifted' from the parent offered element, which may be the root. This is similar to how browsers handle composited elements.
```
khushalsagar, line 94:
```suggestion
When an element is captured, its painting is 'lifted' from the parent offered element, which may be the root. It's analogous to the nested elements having opacity:0 in the parent element's image. This is similar to how browsers handle composited elements.
```

explainer.md
khushalsagar: The idea sounds interesting. Do you have any example in mind that could make a case for your parent being different from the nearest ancestor shared element? And also for having a completely flat list.

Because the alternate choice we were debating was to have every offered element be a 'transition container'. With the assumption that there is no use-case where the developer would want to change this hierarchy or have a flat list. If we could point to a use-case then it would help with making an argument for the flat mode being a useful mode and not just a stop-gap until we have hierarchy.

explainer.md
```diff
@@ -89,21 +89,29 @@ The second mode where styles are copied to a container element won't be part of
...
+### Retaining hierarchy during a transition
+
+By default, offered elements are captured in a flat hierarchy. As in all offered elements, including the root, will be laid out as siblings. The browser uses cached transforms to position each element so it overlaps exactly with its quad on the old page.
+
+The means that, during the transition, scaling one offered element won't impact the rendering of another offered element, even if it was a child of the other element in the old page.
+
+This also allows elements to visually move between containers in viewport space, even if they were clipped to some parent in old page.
+
+Alternatively, the developer can make an offered element a 'transition container'. Offered elements will be nested within their closest transition container, and the cached transform will position the element within that container rather than the viewport. This is similar in spirit to how `position: relative` creates a new 0,0 for absolutely positioned elements.
```
khushalsagar, line 104:
```suggestion
Alternatively, the developer can make an offered element a 'transition container'. Offered elements will be nested within their closest transition container, and the cached transform will position the element within that container rather than the viewport. This is similar in spirit to how `position: relative` becomes the containing block for absolutely positioned elements.
```

explainer.md
```diff
@@ -89,21 +89,29 @@ The second mode where styles are copied to a container element won't be part of
...
+
```
khushalsagar, line 105: We should also mention the clipping and effects being preserved if the developer uses style + image mode when capturing the element. So for example if your ancestor shared element has opacity: 0.5, it won't be preserved if you have a flat list. But with this mode, since the hierarchy is preserved the visual effects continue to be inherited.

explainer.md
khushalsagar, line 114:
```suggestion
The captured elements are displayed using a tree of elements. The root of this tree is a viewport-filling container with fixed position. The offered elements from Page-A are positioned absolutely at (0,0), nested according to their closest 'transition container' or the root if there's no parent container, and moved into their previous viewport-relative (or transition container relative) positions using the cached transform. Their content is painted on top of Page-B which ensures that the user continues to see Page-A's visuals as Page-B is loading. Note that this may not reproduce the exact rendering on Page-A. For example, the relative paint order of shared elements is preserved in this tree rendered on top of Page-B. But if a shared element was occluded by another element, the latter is painted into the root's image unless it is also offered as a shared element.
```

khushalsagar: Using a fixed position root for everything SGTM. We actually also needed that to describe the paint order of transition elements (using z-index). The transition DOM I'm expecting looks something like this :

transition root (position: fixed)
|_ container (position: absolute)
    |_ wrapper
        |_ old image
        |_ new image

Where the transition root has a flat list of containers for each shared element + the root image. If one of the shared elements is offered as the "transition container" for nesting then you'd have something like this :

transition root (position: fixed)
|_ container (position: absolute)
    |_ wrapper
    |   |_ old image
    |   |_ new image
    |_ container (position: absolute)
        |_ wrapper
            |_ old image
            |_ new image

So the container has nested container elements inside it. Does that sound right to you?

explainer.md
khushalsagar: My comment above touches on this DOM structure. Let's discuss it there? I don't think we should need to a new wrapper type but I might be missing the case you had in mind.

explainer.md
ianvollick: Say we have a transition element that's position:fixed -- will its containing block still be the viewport? If that's true, it seems like we could be in a situation where a transition element's transform/positional parent is different than its effect parent. I.e., we'd need something like property trees to represent the relationships and things become far more complex. Could we instead make it be the case that all non-leaf nodes in this transition hierarchy establish a containing block for all positioned descendants?

Also, since this seems a bit involved, could we perhaps add a TODO for clarifying the details of the hierarchical mode and discuss this on a separate, dedicated PR?

explainer.md
```diff
@@ -113,41 +121,60 @@ The CSS images and computed properties/styles cached from Page-A are represented
...
+During the transition a new stacking context (called uber-root) is created with the following hierarchy :
+
+```
+uber-root stacking context
+├─ root stacking context
+└─ transition stacking context
+```
+
+This allows using the output of the root stacking context to provide a live root image for Page-B. The transition elements are painted into the transition stacking context. An alternate approach to this was to paint the transition elements in the [top layer](https://fullscreen.spec.whatwg.org/#top-layer). But that made it difficult to support transitions when there is other content in the top layer (fullscreen elements, dialog) and to ensure effects on the root element which are applied to the root stacking context (background-color, filter) are captured in the root image. See [issue](https://github.com/WICG/shared-element-transitions/issues/74) for detailed discussion.
```
khushalsagar, line 177:
```suggestion
This allows using the output of the root stacking context to provide a live root image for Page-B. The transition stacking context maps to the transition-root element. An alternate approach to this was to paint the transition-root in the [top layer](https://fullscreen.spec.whatwg.org/#top-layer). But that made it difficult to support transitions when there is other content in the top layer (fullscreen elements, dialog) and to ensure effects on the root element which are applied to the root stacking context (background-color, filter) are captured in the root image. See [issue](https://github.com/WICG/shared-element-transitions/issues/74) for detailed discussion.
```

Can clarify what the transition stacking context is if we specify the transition-root element in the DOM structure.

explainer.md
khushalsagar: "we could be in a situation where a transition element's transform/positional parent is different than its effect parent". Do you mean if the developer overrides anything in this DOM to be position:fixed? If we're building this using DOM elements (or pseudo-elements) then why would it be difficult if the positional parent is different than the effect or clip parent. Are you concerned about the "out of process" implementation getting complicated?

ianvollick: > Do you mean if the developer overrides anything in this DOM to be position:fixed?

Yep, that's right.

> If we're building this using DOM elements (or pseudo-elements) then why would it be difficult if the positional parent is different than the effect or clip parent. Are you concerned about the "out of process" implementation getting complicated?

I am, yeah. If engines want to support this outside of renderer processes, than this complexity would need to be replicated.

khushalsagar: The default hierarchy and styles set up by the UA (as described above) should ensure an element's transform parent is the same as its effect/clip parent. This only becomes an issue if the developer can modify the style on these elements and that's allowed for the renderer process implementation (same-origin) only. So it won't impact the out of process implementation.

ianvollick: Could this not be impacted by the styles on transition elements even in the implicit set up for cross-origin? Or are you saying that we'd override the positioning to be `position:absolute` in that case and the only way it would change is if the developer explicitly changed that via JS on the destination page?

khushalsagar: Yes. We're setting the position for all container elements to be position: absolute with a UA stylesheet. Developers can change it only for same-origin via CSS or script.

ianvollick: Got it. Thanks!

explainer.md
jakearchibald: fwiw, I didn't edit this section. I'm not sure I fully understand it.

explainer.md
jakearchibald: Seems ok, but I'm a little worried we're using 'root' a lot to mean different things.

explainer.md
jakearchibald: I tried to include a bit of this in the explainer. A flat list allows you to scale the parent without stretching the children, or needing to provide a counter-transform to provide that. A flat list also makes it easier if a child element moves containers between pages.

explainer.md
jakearchibald: @khushalsagar lol yeah the DOM model I had doesn't make sense. Not sure what I was thinking.

jakearchibald: I reworked the DOM stuff (which meant mostly reverting it back to what it was.

jakearchibald (approved):

PR #102, opened 2021-12-02
Update explainer.md to include resolution of image size.

khushalsagar:


explainer.md
```diff
@@ -60,18 +60,20 @@ The entire painting of the element is captured, including things which appear ou
 
 https://user-images.githubusercontent.com/93594/141118353-d62d19a1-0964-4fa0-880f-bdde656ce899.mp4
 
-The element is captured without the effects (such as opacity and filters) from parent elements. Effects on the element itself are baked into the image. However, the element is captured without transforms, as those transforms are reapplied later.
+The element is captured without the effects (such as opacity and filters) from parent elements. Effects on the element itself are baked into the image. However, the element is captured without transforms, as those transforms are reapplied later. The root is always captured as a single image, with the other transition elements removed (similar to how compositing works today).
```
vmpstr, line 63: > similar to how compositing works today

I'm not sure what that means, maybe remove it?

explainer.md
```diff
@@ -60,18 +60,20 @@ The entire painting of the element is captured, including things which appear ou
...
 
-- [Open question](https://github.com/WICG/shared-element-transitions/issues/72): Should we have a way to expand the root capture area for particular transitions? For example, transitions that involve vertical movement?
-- [Open question](https://github.com/WICG/shared-element-transitions/issues/73): Other elements can also be massive. Do we need a way to limit and control the captured size of those?
+#### Image Size
+The size of the image cached for an element is equal to its border box size + the ink overflow from visual effects like shadows and blurs applied to the element. This allows exposing parts of an element during the transition which may have been hidden earlier. If an element is massive a UA may need to clip the image to a capped size (a common case would be the max texture size supported by the device). When caching a subset of the element due to this constraint, the rect within the element cached by the UA is the area closest to the viewport (computed using a heuristic).
```
vmpstr, line 70:
```suggestion
The size of the image cached for an element is equal to the element's [ink overflow rectangle](https://drafts.csswg.org/css-overflow-3/#ink-overflow-rectangle). This allows exposing parts of an element during the transition which may have been hidden earlier. The user-agent is allowed to clip the image to an implementation defined size (a common case would be the max texture size supported by the device). When caching a subset of the element due to this constraint, the area within the element cached by the user-agent is the area closest to the viewport.
```

(I _think_ what you're describing is the ink overflow rectangle, but please verify)

explainer.md
```diff
@@ -60,18 +60,20 @@ The entire painting of the element is captured, including things which appear ou
...
+
+The size of the root image and the area captured follows a pattern similar to shared elements. However, since the root image is generated using the root stacking context it is likely to exceed and be clipped to capped size in most cases.
```
vmpstr, line 72:
```suggestion
The size of the root image and the area captured follows a pattern similar to shared elements. However, since the root image is generated using the root stacking context it is likely to be clipped to an implementation defined size in most cases.
```

vmpstr (approved):
Mostly wording nits

explainer.md
khushalsagar: This was meant to indicate that removing the elements is similar to how browsers composite elements into separate textures today. The alternate text was something like : "The root is always captured as a single image, with the other transition elements removed (as-if those elements have opacity 0 in the image)."

Is that better? Also @jakearchibald on this.

explainer.md
khushalsagar: I don't think it's ink overflow since the text there mentions : "of a box and its contents". I'm assuming contents implies overflow from the element's descendents in addition to the element's own visual effects. While we only mean overflow from the element's own effects.

khushalsagar: Discussed offline. Saying ink overflow rect here is fine since paint containment already implies that overflow from the content (or descendents) will be clipped.

This does imply that we can have overflow due to overflow-clip-margin and we assumed that the image size doesn't need to expand for that. But we can resolve that detail later.

explainer.md
khushalsagar: We can take this in a separate PR. Wanted to get Jake's thoughts too since he added this.

Issue #64, opened 2021-11-03
Defining end of transition to remove pseudo-elements created for the transition.


khushalsagar:
A point that came up during design discussions in the [explainer](https://github.com/WICG/shared-element-transitions/blob/new-explainer/explainer.md) was clarifying the event that defines the end of the transition and pseudo-elements created for the transition can be removed. The 2 options are :

- Based on a duration. There is a default UA one but it can be overridden by the developer. The timer starts at the first rendering lifecycle update and the transition completes at this duration.
This is easy to reason about and developers can also do rAF driven animations which are not tied to any browser animation framework (CSS transitions, CSS animations or Web Animations).

- Based on whether an element has an active animation (via the browser animation frameworks mentioned above). Once there is no active animation, the transition completes.
This makes it easy to set up browser animations and have the state be cleaned up automatically. But would end with developers setting no-op animations if they are being rAF driven or there is a delay in fetching stylesheets/script which set up the actual animation.

vmpstr:
I'm fine with not "easily supporting" rAF animations for this. So I think doing the clean up when there are no animations is fine. This is somewhat of an implementation detail, so if someone does use a rAF animation they will be unpleasantly surprised which will likely result in a bug being filed. But that being said, we can always switch to a timer based solution if these cases frequently arise

A side question to this, and I know I've brought this up in other contexts before, but do we need a strict cap on the animation duration?

jakearchibald:
When I was looking at the API proposal my intention was:

- Transition is done when all animations within the animation stage are finished. When an animation finishes, we'd allow the events to fire and the promises to resolve, then check to see if there are any non-finished animations. This would allow animations to chain.
- We wouldn't support raf-based animations in v1 (they're not easy to do for pseudo elements anyway), but we would easily add a way for the developer to provide a promise that settles once the animation is done (similar to `waitUntil` in various APIs).
- There'd be no limit on transition length, since the page is just delaying itself.

PR #107, opened 2021-12-03
Fleshing out the API

jakearchibald:


explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
-document.performTransition((transition) => {
-  // Build up transition parts, then…
-  document.documentElement.pseudo("page-transition-container(name)").animate(...);
-});
-```
+The styles to apply these animations will be dynamically added to the UA stylesheet, and removed once the transition is complete.
+
+- Open question: What are the default durations and easings?
+- Open question: Are the generated keyframes usable outside the UA stylesheet? If so, what are they called?
+- Open question: I worry that the default scaling animation makes it hard to do animations where both images cross-fade but don't change size, which feels like a common type of transition. I think this needs more thought.
```
jakearchibald, line 314: I need to think about this and flesh it out some more. In another iteration I'll detail the dynamic CSS generated by the UA in more detail.

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
+  - The browser supports page transitions.
+  - This document is navigating to another (`pagehide` also fires if the iframe/tab is being removed).
+  - The next page is same-origin, although cross-origin may be supported in future.
+  - The window is top-level, although nested page transitions may be supported in future.
+- `event.transition.setElement(element, tag, options)` - Set an element to be used in the transition.
+  - `element` - the element.
+  - `tag` - the tag name. Can be z to un-set this element.
+  - `options` - reserved for future use. This is where "computed style + content image" and "retaining hierarchy" modes will be exposed.
+- `event.transition.setData(data)` - An object that is structured-cloned and passed to the next page.
+- `event.transition.ignoreCSSTaggedElements()`
```
vmpstr, line 234: how are elements combined if this isn't called?

This is one of the bad things about providing an API in either script or JS, we need to figure out what happens if both script and js are there.

vmpstr: nevermind, i see you describe this below -- but maybe comment "see below"

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
 
-This sketch is particularly half-baked. A more concrete proposal will be possible when more of the concepts are decided.
+1. If `ignoreCSSTaggedElements` was not called, gather elements offered via `page-transition-tag`.
+1. Add/remove offered elements according to `setElement` calls.
+1. If at least one element remains offered, a transition can go ahead.
 
-- [Open question](https://github.com/WICG/shared-element-transitions/issues/79): Do we need `createTransitionElement`? It could be done via adding CSS properties, but it might be clumsy if the developer is going to immediately remove the properties afterwards.
-- [Open question](https://github.com/WICG/shared-element-transitions/issues/80): How does the outgoing page offer just the root for transition?
-- [Open question](https://github.com/WICG/shared-element-transitions/issues/81): When is the `performTransition` callback called?
+As a result of the above, if an element is offered via CSS and `setElement`, then `setElement` wins. If there are multiple calls to `setElement` for the same element, the last wins.
```
vmpstr, line 244: To clarify, so we can have multiple elements map to the same tag (in css at least?) but an element can only be mapped to a single tag? Why is this a restriction? And tbh, how can we have the same tag used on multiple elements (I'm guessing this comes later, but I'm commenting as I'm reading)

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
 
-## Defining the animation
+Gathering offered elements _after_ the dispatch of `pagehide` means the developer can use a mix of the CSS and JS methods. For instance, the developer could set a class on the root element depending on `event.nextUrl`, which changes the offered elements.
```
vmpstr, line 246: This implies that we resolve style after pagehide event fires, which means we, for example, will likely run rAF callbacks? I'm just trying to understand if pagehide happens too late in the pipeline (like unload?)

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
 
-How will Page-B define the animation?
+## Page-B
 
-### Default animation
+Page-B must also offer elements to use for a transition to happen. Page-B also controls the animation of the transition.
```
vmpstr, line 250: Both of the "musts" here preclude exit/enter transitions. Is that intentional?

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
-- If an element exists in both Page-A and Page-B, and neither are 'container and child', a transform animation takes its container from Page-A size/transform to Page-B size/transform, while cross-fading the two images.
+#### Offering Page-B elements
 
-- [Open question](https://github.com/WICG/shared-element-transitions/issues/82): What if the Page-A element is 'container and child' but the Page-B element is 'single image'?
+Page-B elements are offered for the transition using the same mechanism as in Page-A, `page-transition-tag`.
 
-Because the images are sized to 100% of the container, the images will also change size throughout the transition. How these are scaled can be changed using regular CSS features like `object-fit`.
+Elements from Page-A and Page-B that have the same tag are merged into one transition elements as documented in "Mixing in elements from Page-B and associating them with transition elements from Page-A". If multiple elements have the same tag, they're associated in DOM order.
```
vmpstr, line 258: Clarifying: if I have 3 elements with tag 'foo' on the old page and 5 on the new page, then the last 2 on the new page are not transitioned?

Also random thought: I'm not sure how this works with shadow dom. I can presumably have shadow dom elements transition here, I guess they also have a well defined DOM order? But is that the same order as the rendered order?

Things like querySelectorAll don't seem to specify the order in which they return the element. querySelector says "first" but I guess all of this is also DOM order?

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
-- `::page-transition-root-container` - Useful for cases where something needs to be rendered underneath the root images.
+#### Default styles
 
-These will be selecting UA generated pseudo-elements.
+These styles will be in the UA stylesheet:
 
 ```css
-::page-transition-container(header) {
-  animation-delay: 300ms;
+::page-transition {
```
vmpstr, line 279: as a nit, that was also not clear to me in Khushal's CL: is this ::page-transition or :root::page-transition?

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
+  position: absolute;
+  inset: 0;
+  isolation: isolate;
+}
+
+::page-transition::image {
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 100%;
```
vmpstr, line 295: that's a 100% of the image-wrapper which is inset 0, what's the image-wrapper's containing block? page-transition?

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
 ```
 
-CSS can be used to make changes to the default animation, or override `animation-name` to remove the default.
+#### Transition animations
 
-### JavaScript access to elements
+The default animations are as follows:
 
-So far the transition elements have been addressed by pseudo-element selectors, but JavaScript could be given access to the elements using the [CSSPseudoElement](https://drafts.csswg.org/css-pseudo-4/#CSSPseudoElement-interface) interface.
+- The `::image(incoming)` animates _from_ opacity 0 with a fill-mode of `both`.
```
vmpstr, line 306: Why not just add these into css block above?

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
+- The `::image(outgoing)` animates _to_ opacity 0 with a fill-mode of `both`.
+- If an element exists in both Page-A and Page-B, a transform animation takes its container from Page-A size/transform to Page-B size/transform.
```
vmpstr, line 308: To clarify: if it only exists in one of the pages, it fades/appears in place?

But it can't happens unless at least one is offered up on both sides? :)

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
-});
-```
+The styles to apply these animations will be dynamically added to the UA stylesheet, and removed once the transition is complete.
+
+- Open question: What are the default durations and easings?
+- Open question: Are the generated keyframes usable outside the UA stylesheet? If so, what are they called?
+- Open question: I worry that the default scaling animation makes it hard to do animations where both images cross-fade but don't change size, which feels like a common type of transition. I think this needs more thought.
+- [Open question](https://github.com/WICG/shared-element-transitions/issues/84): Default animations work well for things which are at least partially in-viewport in both Page-A and Page-B, but it gets tricky if you consider a non-sticky header that scrolled out of view by 1000s of pixels.
+- [Open question](https://github.com/WICG/shared-element-transitions/issues/83): When will the default animation start? When the browser would usually first render Page-B?
+
```
vmpstr, line 317: Another question, unless I missed it, is where are the transform/positioning of the elements specified?

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
+### Via JavaScript
+
+The JavaScript API extends on the capabilities of the CSS API.
 
-- [Open question](https://github.com/WICG/shared-element-transitions/issues/86) What's the deadline for calling `performTransition`?
+```js
+document.addEventListener("beforepageshow", async (event) => {
+  if (!event.transition) return;
+  event.transition.setElement(document.documentElement, "root");
+  event.transition.setElement(document.querySelector(".header"), "header");
```
vmpstr, line 330: I'm not sure if discuss this below, but one problem of this is that if we want to wait for `.header` to at least be parsed before `beforepageshow` fires, but it's unclear if UA can know what are the important elements, since this is only revealed as a part of `beforepageshow` handler

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
+  });
+});
+```
 
-## Signaling the end of a page transition
+`beforepageshow` is a new event which fires just before the page is shown. Unfortunately `pageshow` fires a long after the page is shown, after `window.onload`. The developer will have to add a listener to this somewhere that's executed before the page is shown.
 
-At the start of the transition, the browser could gather all the `Animation`s active on the stage, and assume the animation is complete once all animations finish.
+- `event.previousURL` - URL of the previous page. Is `null` if cross-origin, or if there was no previous page.
+- `event.direction` - `forward` if moving forward in history, `back` is moving back in history, `new` otherwise.
```
vmpstr, line 345: These two seem to imply we can do an enter transition, but my understanding is that the previous page need to offer something up for the transition to happen

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
+- `event.transition` - `null` if Page-A did not offer any elements.
```
vmpstr, line 346: like this :)

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
+- `event.transition.data` - Data set by Page-A via `setData`.
```
vmpstr, line 347: not here that this is null in cross origin

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
-::page-transition-root-outgoing {
-  animation-name: slide-left;
-}
+  transition.setElement(document.querySelector(".new-message"), "new-message");
 
-::page-transition-root-incoming {
-  left: 100%;
-  animation-name: slide-left;
-}
+  await transition.start();
```
vmpstr, line 380: probably don't need to await it

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
+
 # Cross-fading
 
 Cross-fading two DOM elements is currently impossible if both layers feature transparency. This is due to the default composition operation: black with 50% opacity layered over black with 50% opacity becomes black with 75% opacity.
 
-However, the [plus-lighter](https://drafts.fxtf.org/compositing/#porterduffcompositingoperators_plus_lighter) compositing operation does the right thing when isolated to a set of elements whose `opacity` values add to 1. The "image wrapper" is meant to provide this isolation.
+However, the [lighter](https://drafts.fxtf.org/compositing/#porterduffcompositingoperators_lighter) compositing operation does the right thing when isolated to a set of elements whose `opacity` values add to 1. The "image wrapper" is meant to provide this isolation.
 
-Allowing `mix-blend-mode` to be set to `plus-lighter` will enable developers to create real cross-fades between elements for this feature and elsewhere.
+Allowing `mix-blend-mode` to be set to `lighter` will enable developers to create real cross-fades between elements for this feature and elsewhere.
```
vmpstr, line 413: Can you add a link to the proposal to do this

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
+  page-transition-tag: root;
+}
 .header {
   page-transition-tag: header;
-  page-transition-capture: container-and-child;
 }
 ```
 
-Setting a `page-transition-tag` marks this element as a transition element, and the `page-transition-capture` property is used to indicate how it's captured (as a single image vs as a computed style + content image).
+The tag is a [`<custom-ident>`](https://www.w3.org/TR/css-values/#identifier-value). It's recommended to use 'root' to refer to the bottommost element that covers the viewport, which is usually `:root`, but this isn't enforced.
```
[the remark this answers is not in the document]
jakearchibald: Yeah, I've written the API assuming the root is just any other shared element. I can adapt the the proposal if we change our minds there.

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
-    transition.offerElement('header', document.querySelector('.header'), {
-      capture: 'container-and-child',
-    });
-
-    transition.setData({…});
-  });
+document.addEventListener("pagehide", (event) => {
+  if (!event.transition) return;
+  event.transition.setElement(document.documentElement, "root");
+  event.transition.setData(data);
```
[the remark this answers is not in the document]
jakearchibald: Yeah, intentionally, but I'll make that clearer.

```suggestion
  event.transition.setData({ … });
```

Would this work?

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
 });
 ```
 
-Expanding on the capabilities of the CSS properties, a JS API allows the developer to change which parts are offered depending on the destination of the navigation, and the direction of navigation (back vs forward). This builds on top of the [app-history API](https://github.com/WICG/app-history).
+The [`pagehide`](https://developer.mozilla.org/en-US/docs/Web/API/Window/pagehide_event) event is an existing event which fires during the unload of a document. The event type is currently [`PageTransitionEvent`](https://html.spec.whatwg.org/multipage/indices.html#events-2:pagetransitionevent). This proposal will change that to `PageHideTransitionEvent`, which extends `PageTransitionEvent`, and adds the following:
```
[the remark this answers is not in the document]
jakearchibald: Yeah, I'm happy for more discussion on this. I'd rather use an existing event if one fits, but we can make something new if nothing fits.

> Like does this fire while the document is still alive and well? Can it still run lifecycles?

What would it mean to fire an event when a document isn't alive?

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
 
-Also, data can be provided via `setData`. This can be anything structured-clonable, and will be made available to Page-B.
+- `event.nextURL` - The URL of the next page. Is `null` if cross-origin, or if there's no next page for this window.
```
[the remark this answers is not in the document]
jakearchibald: Yeah, I thought "there's no next page" captured that.

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
+- `event.transition` - `null` if a transition cannot be performed. The following needs to be true for transitions to be performed:
+  - The browser supports page transitions.
+  - This document is navigating to another (`pagehide` also fires if the iframe/tab is being removed).
```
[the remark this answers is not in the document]
jakearchibald: Visibility change also fires when changing from one tab to another, or minimising a window, or when moving to another desktop on the OS. Whereas `pagehide` seems more like "navigate away".

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
+  - The next page is same-origin, although cross-origin may be supported in future.
+  - The window is top-level, although nested page transitions may be supported in future.
+- `event.transition.setElement(element, tag, options)` - Set an element to be used in the transition.
+  - `element` - the element.
+  - `tag` - the tag name. Can be z to un-set this element.
```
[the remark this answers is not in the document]
jakearchibald: lol I pressed z when I meant `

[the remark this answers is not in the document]
jakearchibald:
```suggestion
  - `tag` - the tag name. Can be `null` to un-set this element.
```

explainer.md
jakearchibald: I'm not sure what it would mean for an element to have many tags. The tag is like a "transition group", and elements can only be part of one group.

explainer.md
jakearchibald: Maybe it does. We can come up with another event if it's too late.

explainer.md
jakearchibald: I don't think it does. Page-B must offer one element for the transition. What would it look like if it offered none? It would only be Page-A things fading out to… what? A white background?

explainer.md
jakearchibald: > Clarifying: if I have 3 elements with tag 'foo' on the old page and 5 on the new page, then the last 2 on the new page are not transitioned?

I don't think I explained this well. The first three would be matched, but the last 2 would be unmatched but still part of the transition (entry transitions).

jakearchibald: The ordering stuff is interesting. I need to think about this more. Especially if we're still thinking of allowing elements in iframes to be shared elements.

explainer.md
jakearchibald: It's technically `:root:page-transition`, but I don't think it matters

explainer.md
jakearchibald: Yeah, since it's position absolute

explainer.md
jakearchibald: Good question. If we allow the start of the transition to be delayed, we wouldn't want this animation definition to be there at the start.

explainer.md
jakearchibald: > To clarify: if it only exists in one of the pages, it fades/appears in place?

Yeah, I'll clarify that

explainer.md
jakearchibald: Dynamically inserted in the UA stylesheet. Yeah I need to define this further, but how it works with keyframes isn't clear to me yet. I've got some ideas around this which I'll jot down.

explainer.md
jakearchibald: Yeah, eventually we'll need a way to wait for stuff. Question is if we need it for v1.

explainer.md
jakearchibald: Not sure I understand

explainer.md
jakearchibald: It depends when the elements are created. Happy to make it sync if possible.

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
+Multiple elements can share a tag. For example, each comment of an article can use `page-transition-tag: comment`, and each comment will be offered for the transition.
 
-Elements on Page-A and Page-B which have the same `page-transition-tag` are automatically associated with each other. If multiple elements have the same value, they're associated in DOM order.
+The other modes mentioned in this document, such as "computed style + content image", and "retaining hierarchy" will be exposed via other CSS properties, and `page-transition` will be used as a shorthand. However, this is out of scope for 'v1'.
 
-Using CSS properties means values can change depending on viewport (`@media`) and browser support (`@supports`).
+### Via JavaScript
 
-### JS API
+The JavaScript API extends on the capabilities of the CSS API. In particular, it allows the developer to react to the URL of Page-B.
```
khushalsagar, line 211: It's not just about reacting to the URL of page-B. It's also about the different states page-A can be in when the transition occurs. For example a transition after page-A's initial load vs after the user has interacted with page-A which might have changed its content.

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
+- `::page-transition(tag)` - Select the 'transition element' of a given `page-transition-tag`.
+- `::page-transition(tag)::image-wrapper` - Select the 'image wrapper'.
+- `::page-transition(tag)::image(outgoing | incoming)` - Select the incoming or outgoing image.
 
-```css
-::page-transition-container(header) {
-  /* … */
-}
-```
+`::page-transition` without a tag selects all transition elements.
```
khushalsagar, line 270: Could we use ::page-transition(*) for this instead? Mostly because I don't know if we have precedent for this. :) My reading of the parser code was that we either have selectors with no arguments or with arguments.

explainer.md
khushalsagar: I ended up adding the animation styles when we hit the "start" phase of the transition. I'm assuming that'll come up later in this proposal.

explainer.md
khushalsagar: This could be a UA detail but we'll need to do a width/height animation instead of changing size through a transform animation if the aspect ratio is different between the 2 images and we don't want the content to stretch as this ratio is changing.

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
-// In Page-B
-document.performTransition((transition) => {
-  // Build up transition parts, then…
-  document.documentElement.pseudo("page-transition-container(name)").animate(...);
-});
-```
+The styles to apply these animations will be dynamically added to the UA stylesheet, and removed once the transition is complete.
+
+- Open question: What are the default durations and easings?
+- Open question: Are the generated keyframes usable outside the UA stylesheet? If so, what are they called?
```
khushalsagar, line 313: We could add a keyword which indicates the default keyframe (if any) generated for each of the new pseudo elements.

We'll need to expose this in script too (to be usable with Web Animations).

explainer.md
khushalsagar: I didn't follow how the default scaling animation makes this hard. It'll be a no-op if the size doesn't change.

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
+- Open question: I worry that the default scaling animation makes it hard to do animations where both images cross-fade but don't change size, which feels like a common type of transition. I think this needs more thought.
+- [Open question](https://github.com/WICG/shared-element-transitions/issues/84): Default animations work well for things which are at least partially in-viewport in both Page-A and Page-B, but it gets tricky if you consider a non-sticky header that scrolled out of view by 1000s of pixels.
+- [Open question](https://github.com/WICG/shared-element-transitions/issues/83): When will the default animation start? When the browser would usually first render Page-B?
```
khushalsagar, line 316: We need to define both : when we start looking for elements with a transition-tag on the new page and start the animation.

And yeah, we should do it on page-B's first render with render-blocking helping developers control what blocks the first render.

explainer.md
khushalsagar: > how it works with keyframes isn't clear to me yet

Happy to help clarify this.

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
+
+All of these can be overridden by the developer via CSS animations.
+
+The animation is considered complete once all animations on all the pseudo-elements finish.
```
khushalsagar, line 320: This actually ties into the open question I had here : https://github.com/WICG/shared-element-transitions/issues/64.

I was hoping this could be explicit about animations which are added on page-B's first render. Because otherwise we have to keep track of new animations getting added/removed while a transition is active.

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
+
+### Via JavaScript
+
+The JavaScript API extends on the capabilities of the CSS API.
 
-- [Open question](https://github.com/WICG/shared-element-transitions/issues/86) What's the deadline for calling `performTransition`?
+```js
+document.addEventListener("beforepageshow", async (event) => {
```
khushalsagar, line 327: beforepageshow is fired before the first raf, right?

explainer.md
khushalsagar: That problem is going to be an issue for the purely declarative stuff too, we wouldn't know if we've fetched/parsed enough of the document for all shared elements to be discovered. It's better to tackle this as a separate problem (like render-blocking) such that it can be used with transitions later. Until then, we can rely on UA heuristics for this.

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
+  if (!event.transition) return;
+  event.transition.setElement(document.documentElement, "root");
+  event.transition.setElement(document.querySelector(".header"), "header");
 
-Note that using pseudo-elements here implies that the developer can not change the DOM structure of transition elements. An alternate approach considered to provide this extensibility uses shadow DOM. The pseudo-element option was preferred to make it easier to build on concepts which already exist in the platform and ensure ease of implementation. See [issue](https://github.com/WICG/shared-element-transitions/issues/93) for detailed discussion.
+  await event.transition.ready;
```
khushalsagar, line 332: What is "ready" used for?

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
+- `event.transition.abandon()` - Abandon the transition.
+- `event.transition.setElement(element, tag, options)` - Same API as in Page-A.
+- `event.transition.ignoreCSSTaggedElements()` - Same API as in Page-A.
 
-In addition, the JS API could include a way to override this by letting the developer provide a promise which keeps the transition active, allowing for animations driven some other way, such as `requestAnimationFrame`.
+In future, this API could include:
 
-This is discussed in [#64](https://github.com/WICG/shared-element-transitions/issues/64).
+- A way to delay the start of the transition, eg to wait for an image to load.
+- A way to enumerate the pseudo elements, which will depend on [CSSPseudoElement](https://drafts.csswg.org/css-pseudo-4/#CSSPseudoElement-interface).
```
khushalsagar, line 355: Didn't follow this.

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
+- A way to manually associate Page-A and Page-B elements, to change the default.
```
khushalsagar, line 356: And this too. The associating of elements can already be controlled by the developer via tag names?

explainer.md
khushalsagar: Oh I see, is that what "ready" is for in the MPA version? I do like the idea of a "finalize" signal. The developer indicating that all shared elements on the new page have been specified and elements can now be paired.

But that can just be a sync API, so transition.start() can be sync.

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
-Also, data can be provided via `setData`. This can be anything structured-clonable, and will be made available to Page-B.
+- `event.nextURL` - The URL of the next page. Is `null` if cross-origin, or if there's no next page for this window.
+- `event.transition` - `null` if a transition cannot be performed. The following needs to be true for transitions to be performed:
+  - The browser supports page transitions.
+  - This document is navigating to another (`pagehide` also fires if the iframe/tab is being removed).
+  - The next page is same-origin, although cross-origin may be supported in future.
+  - The window is top-level, although nested page transitions may be supported in future.
+- `event.transition.setElement(element, tag, options)` - Set an element to be used in the transition.
+  - `element` - the element.
+  - `tag` - the tag name. Can be `null` to un-set this element.
```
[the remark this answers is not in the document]
jakearchibald: Yep

explainer.md
jakearchibald: `::page-transition(*)` seems fine to me!

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
 
-```css
-::page-transition-container(header) {
-  /* … */
-}
-```
+`::page-transition` without a tag selects all transition elements.
 
-Element selectors:
+`::page-transition(tag nth odd)` can be used to select a subset of elements with the same tag, where 'odd' can be any valid [An+B](https://www.w3.org/TR/css-syntax-3/#anb-microsyntax).
```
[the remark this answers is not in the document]
jakearchibald: Yeah. Let's explore dropping the multi-tag thing. I know I was pushing for that, but it seems to be causing more problems than it solves.

explainer.md
jakearchibald: It definitely isn't a UA detail 😄 . Some thoughts on this https://github.com/WICG/shared-element-transitions/issues/108

explainer.md
khushalsagar: Hmmm, then I'm not understanding the case. The text here says, "both images cross-fade but don't change size".

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
-.share-button {
-  page-transition-tag: share-button;
-}
+  await transition.captureAndHold();
 
-/* Slide the roots from right to left */
-@keyframes slide-left {
-  to { transform: translateX(-100%); }
-}
+  await coolFramework.changeTheDOMToPageB();
```
khushalsagar, line 376: We could have a transition.abort() API as well. It's there is the current API (as abortSignal) so you can call prepare but decide to abort the transition.

I was just chatting with Callie about how they're using the abortSignal. They have a use-case where they could either transition to the actual page (if it loads in time) or a loader screen. And in the latter case they abort the transition. With the new API now, the abort could be implicit (for example remove animations on all pseudo elements in a developer stylesheet). But an API for the developer here would make this easier.

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
-- [Open question](https://github.com/WICG/shared-element-transitions/issues/84): Default animations work well for things which are at least partially in-viewport in both Page-A and Page-B, but it gets tricky if you consider a non-sticky header that scrolled out of view by 1000s of pixels.
-- [Open question](https://github.com/WICG/shared-element-transitions/issues/85): If the developer wants a default animation of the root only, how do they define that?
+The pseudo-elements are constructed as documented in "how are transition elements represented".
 
-### CSS animation
+These pseudo-element selectors provide access to these pseudo-elements via `:root`:
 
-CSS can be used to build on default animations, or override the default.
+- `::page-transition(tag)` - Select the 'transition element' of a given `page-transition-tag`.
+- `::page-transition(tag)::image-wrapper` - Select the 'image wrapper'.
```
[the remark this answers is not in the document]
jakearchibald: I agree with you, I'll go with `::page-transition-image-wrapper(tag)`.

explainer.md
```diff
@@ -186,225 +181,236 @@ https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b
...
+- `event.transition` - `null` if a transition cannot be performed. The following needs to be true for transitions to be performed:
+  - The browser supports page transitions.
+  - This document is navigating to another (`pagehide` also fires if the iframe/tab is being removed).
+  - The next page is same-origin, although cross-origin may be supported in future.
+  - The window is top-level, although nested page transitions may be supported in future.
+- `event.transition.setElement(element, tag, options)` - Set an element to be used in the transition.
+  - `element` - the element.
+  - `tag` - the tag name. Can be z to un-set this element.
+  - `options` - reserved for future use. This is where "computed style + content image" and "retaining hierarchy" modes will be exposed.
+- `event.transition.setData(data)` - An object that is structured-cloned and passed to the next page.
```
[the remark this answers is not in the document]
jakearchibald: I'm not going to add this, as it might suggest that other things may be supported for cross-origin cases, meaning we'll have to cover the doc in qualifications like "maybe not for cross-origin!"

The doc already contains:

> Cross-origin transitions are something we want to tackle, but may have significant differences and restrictions for security reasons. Cross-origin transitions are not covered in this document.

…at the start, and:

> Cross-origin transitions aren't yet defined, but are likely to be heavily restricted.

…at the end.

I'll add some other statements to make it clear that a transition will not happen cross-origin.

explainer.md
jakearchibald: Yeah

explainer.md
jakearchibald: The pseudo elements to be created and be in a renderable state.

explainer.md
jakearchibald: As a developer, it isn't clear how I can tell which elements the previous pages has offered.

explainer.md
jakearchibald: `AbortSignal` sounds right.

vmpstr (approved):
Let's target to merge soon

jakearchibald:
Done!

PR #114, opened 2022-01-04
Clarify when pseudo-elements are removed.

khushalsagar:


jakearchibald:
Do we want to be a little more specific about this to allow for the finish of one animation to trigger another animation? We could check at the end of the render steps, or wait for two frames with no active animations.

khushalsagar:
We do wait for at least one rendering lifecycle update/frame to allow the developer to set up an animation. After that you can chain them but there should be an active animation on at least one pseudo element every frame.

vmpstr:
We can wordsmith a bit more, but I think it's worthwhile to merge so that there is at least some indication of when we end the transition

PR #149, opened 2022-05-10
Draft of developer guide

jakearchibald:


developer-guide.md
```diff
@@ -0,0 +1,353 @@
...
+  }
+
+  // With a transition
+  const transition = document.createDocumentTransition();
+  await transition.start(() => updateDOMForPage(data));
+  console.log("Transition complete!");
+}
+```
+
+Where `updateTheDOMSomehow` is a function you'd write to switch the DOM from the current state to the new state.
```
khushalsagar, line 36: Did you mean updateDOMForPage instead of `updateTheDOMSomehow`?

developer-guide.md
```diff
@@ -0,0 +1,353 @@
...
+
+## Async DOM updates
+
+```js
+const transition = document.createDocumentTransition();
+await transition.start(callback);
+console.log("Transition complete!");
+```
+
+The API supports asynchronous DOM updates, which are common in most frameworks. To make this work, `callback` should return a promise that resolves once the DOM update is complete. However, remember that the user is left with a non-interactive screenshot between `callback` being called and its promise resolving, so this should happen as fast as position. Specifically, avoid things like network activity – do those before creating the transition, like this:
```
khushalsagar, line 162: "as fast as possible"

developer-guide.md
```diff
@@ -0,0 +1,353 @@
...
+
+Although, it's possible to have one of these structures that doesn't have an incoming/outgoing image. For example, if the element was only there before the DOM change, or only there after the DOM change.
+
+In a similar way that we were using `::page-transition-outgoing-image(root)` to target the outgoing image of the root, we can use `::page-transition-outgoing-image(site-header)` to target the outgoing image of the site-header, and animate it separately.
+
+The default animation animates the `width` and `height` of the `::page-transition-container` from its before size to its after size, while animating a `transform` to move it from its before position to its after position. If the header is unchanged, it'll appear fixed in place while the rest of the content (the root) slides as we specified earlier.
+
+### Flat transition structures
+
+Each captured element, along with the root, will be a direct child of the special top layer. This means you can move elements between containers, even if they have `overflow: hidden` or some other form of clipping in the real DOM.
```
khushalsagar, line 231: "Each captured element, along with the root, generates a subtree of pseudo elements which is direct child of the special top layer"

Otherwise it sounded like we're saying the element will be attached to the top layer.

developer-guide.md
```diff
@@ -0,0 +1,353 @@
...
+
+We plan to add a feature to allow one transition container to be nested within another, but that isn't currently implemented.
+
+## Animation synchronization gotcha
+
+Right now, there's a delay between Chrome creating the outgoing image element, and the incoming image element. That means the styles for the outgoing image element will apply sooner, which in turn means its animation will start sooner, and apply out of sync with the incoming image element.
+
+This is considered a bug (well, a design error), but you can work around it:
```
khushalsagar, line 239: Sounds like we should make crbug.com/1310798 OT blocking? :)

developer-guide.md
```diff
@@ -0,0 +1,353 @@
...
+}
+```
+
+This pauses the animations until both the outgoing and incoming content is ready.
+
+## Using `object-fit` and `object-position`
+
+The default animation animates the `width` and `height` of the `::page-transition-container`, which is generally frowned upon in web performance circles, as it runs layout per frame. However, for page transitions, we plan to optimize it so it isn't an issue (it isn't optimized yet).
+
+To benefit of this, is `object-fit` and `object-position` become really useful.
```
khushalsagar, line 280: "The benefit of this"?

developer-guide.md
```diff
@@ -0,0 +1,353 @@
...
+.transition-warming-up::page-transition-outgoing-image(*) {
+  animation-play-state: paused !important;
+}
+```
+
+This pauses the animations until both the outgoing and incoming content is ready.
+
+## Using `object-fit` and `object-position`
+
+The default animation animates the `width` and `height` of the `::page-transition-container`, which is generally frowned upon in web performance circles, as it runs layout per frame. However, for page transitions, we plan to optimize it so it isn't an issue (it isn't optimized yet).
```
khushalsagar, line 278: Not sure if we need to mention: "which is generally frowned upon in web performance circles, as it runs layout per frame. However, for page transitions, we plan to optimize it so it isn't an issue (it isn't optimized yet)."

I thought its frowned upon coz running layout on complex trees is slow but in this case it's about running on a thread with possibly lots of contention. We could clarify but it wasn't clear if this detail is needed here.

developer-guide.md
```diff
@@ -0,0 +1,353 @@
...
+  /* As the container transitions from 4:3 to 16:9, keep the outgoing image in the center,
+  revealing blank space on the sides. */
+  object-fit: contain;
+}
+
+::page-transition-incoming-image(lightbox-image) {
+  /* Don't fade the image in, allow it to be shown straight away, but under
+  the outgoing image (the outgoing image will quickly fade out) */
+  animation: none;
+  /* As the container transitions from 4:3 to 16:9, gradually un-crop the image. */
```
khushalsagar, line 302: We're soon going to add `overflow:visible` to the UA css to show shadows/overflow. That will also make it so cover doesn't clip anymore. Could we add `overflow:clip` to dev CSS explicitly make sure this clips?

What this means though is that until box decorations can be captured as styles (rather than in the screenshot) you can't get the clipping cover has.

developer-guide.md
```diff
@@ -0,0 +1,353 @@
...
+
+### Flat transition structures
+
+Each captured element, along with the root, will be a direct child of the special top layer. This means you can move elements between containers, even if they have `overflow: hidden` or some other form of clipping in the real DOM.
+
+We plan to add a feature to allow one transition container to be nested within another, but that isn't currently implemented.
+
+## Animation synchronization gotcha
+
+Right now, there's a delay between Chrome creating the outgoing image element, and the incoming image element. That means the styles for the outgoing image element will apply sooner, which in turn means its animation will start sooner, and apply out of sync with the incoming image element.
```
vmpstr, line 237: Can we maybe reference github issues for things like this, so people can chime in if they have opinions?

developer-guide.md
```diff
@@ -0,0 +1,353 @@
...
+  // the real DOM underneath.
+}
+```
+
+The default transition is a cross-fade from the screenshotted state to the new state.
+
+The 'screenshot' is actually a DOM structure created out of pseudo-elements that sits in a special top-layer:
+
+```html
+<special-top-layer>
```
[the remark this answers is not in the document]
jakearchibald: It isn't the literal top layout though is it? Happy to change it though

developer-guide.md
jakearchibald: Hah, good catch!

developer-guide.md
```diff
@@ -0,0 +1,353 @@
...
+const transition = document.createDocumentTransition();
+await transition.start(callback);
+console.log("Transition complete!");
+```
+
+The API supports asynchronous DOM updates, which are common in most frameworks. To make this work, `callback` should return a promise that resolves once the DOM update is complete. However, remember that the user is left with a non-interactive screenshot between `callback` being called and its promise resolving, so this should happen as fast as position. Specifically, avoid things like network activity – do those before creating the transition, like this:
+
+```js
+async function spaNavigate(path) {
+  const data = await fetchDataForPage(path);
```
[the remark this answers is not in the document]
jakearchibald: I'll rename this function to suggest more can happen

developer-guide.md
jakearchibald: It's workaroundble at least, but yeah it'd be good to have it fixed.

developer-guide.md
jakearchibald: I'll add a link

developer-guide.md
jakearchibald: I think if we don't mention it, some smartarse will be on Twitter saying "hahaha Google so dumb, they don't even follow their own performance advice"

jakearchibald:
Landing this so it's there for the talk, but we can continue to tweak it

PR #166, opened 2022-07-13
Create debugging_overflow_on_images.md

khushalsagar:


debugging_overflow_on_images.md
```diff
@@ -0,0 +1,42 @@
...
+```css
+img {
+/* This inherits all CSS values including `overflow` */
+all: inherit;
+}
+```
+
+If updating the CSS is not trivial, a trivial fix can be to add the following inline style to the element.
+
+```html
```
khushalsagar, line 40: In some cases, it looked like the html/css which was affected was from a third-party. A get 'em all option can be to use MutationObserver and add the inline style using JS but that seems hacky so I'm not suggesting that.

debugging_overflow_on_images.md
```diff
@@ -0,0 +1,42 @@
+# Introduction
+
+Starting with Chrome 106, the `overflow` property is applied on the following elements: img, video and canvas. In earlier versions of Chrome, this property was ignored on these elements.
+
+This implies that an image which was earlier clipped to its content box can now draw outside those bounds. Common examples of CSS which can result in this are:
```
chrishtr, line 5:
```suggestion
This implies that an image which was earlier clipped to its content box can now draw outside those bounds. Examples of CSS which can result in this are:
```

chrishtr, line 5:
```suggestion
This implies that an image which was earlier clipped to its content box can now draw outside those bounds if specified to do so in a style sheet. The default browser style sheet applies the following default style to these elements:

```
<content here>
```


Common examples of CSS which can result in this are:
```

debugging_overflow_on_images.md
```diff
@@ -0,0 +1,42 @@
...
+```css
+img {
+  /* object-fit causes the image to scale and fill the complete box.
+     If the aspect ratio is different, the image will draw outside. */
+  object-fit: cover;
+  overflow: visible;
+}
+```
+
+For example, if your page has the following CSS:
```
chrishtr, line 16: Isn't this two different examples? (The one above and the one below.)

debugging_overflow_on_images.md
```diff
@@ -0,0 +1,42 @@
...
+  /* border-radius should cause the image to draw as a circle.
+     But because 'overflow' is set to visible, it is no longer clipped. */
+  border-radius: 50% 50%;
+  overflow: visible;
+}
+```
+
+# Solution
+
+If overridng the `overflow` property to `visible` is not intentional, inspect the CSS applied to the element via [devtools](https://developer.chrome.com/docs/devtools/css/#view). This should allow you to identify the CSS declaration whih is overriding the `overflow` property to `visible` and update it. Note that another common pattern that can update the `overflow` property is:
```
chrishtr, line 29:
```suggestion
If overridng the `overflow` property to `visible` is not intentional, inspect the CSS applied to the element via [devtools](https://developer.chrome.com/docs/devtools/css/#view). This should allow you to identify the CSS declaration which is overriding the `overflow` property to `visible` and update it to `clip` via an additional style sheet rule such as `#myImage { overflow:clip}`. Note that another common pattern that can inadvertently cause `overflow` to be `visible` is:
```

debugging_overflow_on_images.md
```diff
@@ -0,0 +1,42 @@
+# Introduction
+
+Starting with Chrome 106, the `overflow` property is applied on the following elements: img, video and canvas. In earlier versions of Chrome, this property was ignored on these elements.
```
vmpstr, line 3:
```suggestion
Starting with Chrome 106, the following elements respect the `overflow` property: img, video and canvas. In earlier versions of Chrome, this property was ignored on these elements.
```

debugging_overflow_on_images.md
vmpstr, line 5:
```suggestion
This means that an image which was earlier clipped to its content box can now draw outside those bounds. Common examples of CSS which can result in this are:
```

debugging_overflow_on_images.md
```diff
@@ -0,0 +1,42 @@
...
+If overridng the `overflow` property to `visible` is not intentional, inspect the CSS applied to the element via [devtools](https://developer.chrome.com/docs/devtools/css/#view). This should allow you to identify the CSS declaration whih is overriding the `overflow` property to `visible` and update it. Note that another common pattern that can update the `overflow` property is:
+
+```css
+img {
+/* This inherits all CSS values including `overflow` */
+all: inherit;
+}
+```
+
+If updating the CSS is not trivial, a trivial fix can be to add the following inline style to the element.
```
vmpstr, line 38:
```suggestion
If updating the CSS is not trivial, another fix is to add the following inline style to the element:
```

debugging_overflow_on_images.md
```diff
@@ -0,0 +1,42 @@
...
+
+```html
+<img style="overflow:clip !important"></img>
```
vmpstr, line 41:
```suggestion
<img style="overflow:clip"></img>
```

inline style has highest precedence, so you don't need !important here

debugging_overflow_on_images.md
khushalsagar: Done.

debugging_overflow_on_images.md
khushalsagar: Done.

debugging_overflow_on_images.md
khushalsagar: There were some sites where the style applied to an element from a class declaration has !important for the overflow property and it looked like this was third party code. This is why I'm suggesting inline style with !important.

debugging_overflow_on_images.md
khushalsagar: Yup, my bad.  That's why the starting line was "Common examples". These 2 were the breakages from the sites I inspected so explicitly mentioning both to help with debugging.

vmpstr (approved):

PR #172, opened 2022-07-29
Create shared_element_transition.d.ts file

tbondwilkinson:
There were a few methods that I saw in the browser implementation that I thought may need to also be included, but I didn't see any mention of them in the explainer so I left them out. The one exception I made was abandon(), which seemed like that was probably implemented and could be used by users.

I also couldn't find an up to date spec to confirm, [this one](https://tabatkins.github.io/specs/css-shared-element-transitions/) seems out of date.

But let me know if there are other methods or properties (setElement, ignoreCSSTaggedElements, and I didn't see an ongoingTransition global variable anywhere) that you want included in this d.ts file.

### DIFF
--- a/shared_element_transition.d.ts
+++ b/shared_element_transition.d.ts
@@ -1,13 +1,12 @@
 interface DocumentTransition {
-    start(callback: () => Promise<void> | void): Promise<void>;
-    abandon(): void;
-  }
-  
-  interface Document {
-    createDocumentTransition(): DocumentTransition;
-  }
-  
-  interface CSSStyleDeclaration {
-    pageTransitionTag: string;
-  }
-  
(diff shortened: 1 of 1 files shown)


tbondwilkinson:
Another thing to note is that DocumentTransition could use `declare class` instead of `interface`, as it is technically a class not an interface, see for example the [Navigation API](https://github.com/WICG/navigation-api/blob/main/navigation_api.d.ts) where the events and transition are all classes.

Now that I'm typing this out, I'm convincing myself it should be a class, so I'll make that change.

vmpstr:
Sorry for the late reply and thanks for doing this. The API shape has changed a bit since the initial version and is going to change again in the near future (@jakearchibald is working on updating this). Not sure if we want to merge this and update or update it once the final version of the API is decided on.

PR #179, opened 2022-08-11
Explainer rewrite

jakearchibald:
This rewrite removes some of the fine details that are covered in the spec, and attempts to provide more of an overview of the feature.

In particular, I tried to head of some of the questions I expect TAG will ask.

I removed the section about transitions that capture an element's full computed style. It feels like it's too early to devote a lot of detail to that, particularly as it'll animate on the main thread.


explainer.md
```diff
@@ -1,751 +1,423 @@
...
+On Single Page Apps (SPAs) developers can use animation tools such as [CSS transitions](https://developer.mozilla.org/docs/Web/CSS/CSS_Transitions/Using_CSS_transitions), [CSS animations](https://developer.mozilla.org/docs/Web/CSS/CSS_Animations/Using_CSS_animations), and the [Web Animation API](https://developer.mozilla.org/docs/Web/API/Web_Animations_API/Using_the_Web_Animations_API) to create transitions. However, it's not straight forward, mostly due to the period where both states need to exist in the DOM at the same time, so both can be partially visible (eg in a cross-fade). This creates issues with:
 
-In this mode the computed styles of the element are copied over, so they can be
-re-rendered beyond just transforming an image.
+- **Accessibility**: Having both states in the DOM at the same time can create a confusing experience for screen reader users. Juggling the DOM around for the sake of a transition doesn't play well with things like ARIA live regions.
+- **Usability**: The old DOM will continue to exist during the transition-away, leading to complications if the user manages to interact with that content (eg clicking buttons).
```
khushalsagar, line 10:
```suggestion
- **Usability**: The old DOM will continue to exist during the transition, leading to complications if the user manages to interact with that content (eg clicking buttons).
```

explainer.md
```diff
@@ -1,751 +1,423 @@
...
+- The layout dimensions.
+- A CSS transform that would place element from the layout viewport origin to its current quad.
+- An [`object-view-box`](https://drafts.csswg.org/css-images-4/#propdef-object-view-box) that coincides with the element's border box.
 
-- [Open question](https://github.com/WICG/shared-element-transitions/issues/23):
-  How should the default UA animation order these elements? And also handle a
-  change in associated elements between the 2 pages. Currently they are ordered
-  by paint order. For instance, parents are ordered before children, and
-  negative z-index elements are ordered before positive z-index elements.
+The above is also captured for the incoming state, although the image of the element is 'live', as in it's connected to the representation currently in the DOM.
```
khushalsagar, line 77:
```suggestion
The above is also captured for the incoming state, although the image and computed layout properties of the element are 'live', as in it's connected to the representation currently in the DOM.
```

explainer.md
```diff
@@ -1,751 +1,423 @@
...
-image for Page-B. The transition stacking context maps to the transition-root
-element. An alternate approach to this was to paint the transition-root in the
-[top layer](https://fullscreen.spec.whatwg.org/#top-layer). But that made it
-difficult to support transitions when there is other content in the top layer
-(fullscreen elements, dialog) and to ensure effects on the root element which
-are applied to the root stacking context (background-color, filter) are captured
-in the root image. See
-[issue](https://github.com/WICG/shared-element-transitions/issues/74) for
-detailed discussion.
+Where `name` is the `page-transition-tag` value, which is `root` for the root. If the transitioning element only existed on one side of the DOM change, then either the outgoing or incoming image will be missing.
```
khushalsagar, line 92: Should we mention that tagging the root element is configurable. Its implicitly captured using the following UA style rule:

```css
html {
  page-transition-tag: root;
}
```

But developers can rename it or reset it to none.

explainer.md
```diff
@@ -1,751 +1,423 @@
...
+
+These trees are inserted into a `::page-transition` pseudo element, according to the render-order of their transition elements.
```
khushalsagar, line 94:
```suggestion
These trees are inserted into a `::page-transition` pseudo element, according to the paint order of their transition elements.
```

Its a nit since we've used the word paint order in all other documentation.

explainer.md
```diff
@@ -1,751 +1,423 @@
...
+
+The `::page-transition` is rendered in a top-level, filling the viewport.
```
khushalsagar, line 96:
```suggestion
The `::page-transition` is rendered in a top-level stacking context, filling the viewport.
```

I assumed you meant to write something after top-level, not sure if this was it.

explainer.md
```diff
@@ -1,751 +1,423 @@
...
+
+Once this is in place, rendering is resumed.
+
+# Default styles & animation
+
+## `::page-transition-container(*)`
 
-## Part 3: The transition
+Default styles:
```
khushalsagar, line 104: This section would read easier if we just put the style declaration instead of the text. Example for the container:

```css
html::page-transition-container(foo) {
  position: absolute;

  /* Width and height of the incoming element, if any, otherwise the outgoing element. */
  width: 100px;
  height: 100px;

  /* A transform that places it in the viewport position of the incoming element, if any, otherwise the outgoing element */
  transform: translate(100px, 0);
}
```

explainer.md
```diff
@@ -1,751 +1,423 @@
...
 
-Everything is now in place to perform the transition. The developer can animate
-the transition elements created by the UA using the usual APIs, such as CSS and
-web animations.
+- Absolutely positioned to the top left of the parent.
+- Width and height of the outgoing element.
```
khushalsagar, line 107: Since as the spec we create the pseudo DOM only when both the incoming and outgoing state are ready, all the computed properties correspond to the incoming elements if we have any.

explainer.md
```diff
@@ -1,751 +1,423 @@
...
+- A transform that places it in the viewport position of the outgoing element.
 
-https://user-images.githubusercontent.com/93594/141100217-ba1fa157-cd79-4a9d-b3b4-67484d3c7dbf.mp4
+Default animation:
 
-Note that the browser defers displaying elements from Page-B and starting the
-animation until Page-B is ready for first render. This is currently driven by
-internal browser heuristics and is being standardized in the proposal
-[here](https://github.com/whatwg/html/issues/7131).
+- Width and height animates to the dimensions of the incoming element.
```
khushalsagar, line 112: Here too, the precise CSS animation being created by the browser makes it clearer. One of the confusion that Tab had during the spec review was whether we were using WebAnimations for the default animations.

explainer.md
```diff
@@ -1,751 +1,423 @@
...
+
+## `::page-transition-incoming-image(*)`
+
+As above, but animates from opacity 0 to 1.
+
+# Ending the transition
+
+The transition ends once a state is reached where no animations are running on any of the transition pseuduo-elements.
+
+Once this happens, the pseudo-elements are removed, and the real DOM underneath is shown.
```
khushalsagar, line 151: We should clarify somewhere before this that the tagged DOM elements are not painted in the real DOM while they are being displayed via pseudo-elements during the transition. The effect is similar to the elements and their 
descendants having `visibility: hidden` and `pointer-events: none`.

explainer.md
```diff
@@ -1,751 +1,423 @@
...
+_The above shows the parts of the page captured before and after the DOM change._
 
-Transition elements don't need to be associated with another transition
-elements, which allows for transitions involving elements that are only in
-Page-A or only in Page-B.
+For the outgoing state, the following is captured for each part:
 
-Note that the order in which the transition elements are painted can be
-configured by UA and/or developer stylesheets using z-index.
+- An image of the element, including ink overflow if any. The browser may crop this to make efficient use of memory, and that cropping should be biased towards the viewport.
```
khushalsagar, line 72:
```suggestion
- An image of the element, including ink overflow if any. The browser may crop this to make efficient use of memory, and that cropping should be biased towards the viewport. This image also excludes painting of any descendants with a page-transition-tag, as if they have `visibility: hidden`.
```

explainer.md
```diff
@@ -1,751 +1,423 @@
...
-#### Targeting transition elements
+As long as the framework provides a notification when the DOM is updated, which they already do to allow custom handling of elements, the transition API can be made to work with the framework.
 
-The pseudo-elements are constructed as documented in "how are transition
-elements represented".
+# Animating with JavaScript
 
-These pseudo-element selectors provide access to these pseudo-elements via
-`:root`:
+The promise returned by `prepare()` fulfills when both states have been captures and the pseudo-element tree has been successfully built. This provides developers with a point where they can animate those pseudo-elements with the [Web Animation API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Animations_API).
```
khushalsagar, line 233:
```suggestion
The promise returned by `prepare()` fulfills when both states have been captured and the pseudo-element tree has been successfully built. This provides developers with a point where they can animate those pseudo-elements with the [Web Animation API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Animations_API).
```

explainer.md
```diff
@@ -1,751 +1,423 @@
...
-- `::page-transition-outgoing-image(tag)` - Select the outgoing image.
+https://user-images.githubusercontent.com/93594/184120371-678f58b3-d1f9-465b-978f-ee5eab73d120.mp4
 
-Using '\*' instead of a tag selects the equivalent element for every tag.
+This example animates a circular `clip-path` from the position of the click to a size that covers the viewport. Since this isn't possible with CSS alone ([although it may get easier in future](https://github.com/w3c/csswg-drafts/issues/824#issuecomment-1204467456)), the Web Animation API is used.
 
-Issue: It would be nice to be consistent with the '\*' CSS selector in that we
-would like `::page-transition(*)` to have a lower specificity than
-`::page-transition(tag)`. 
+# Customizing the transition based on the type of navigation
```
khushalsagar, line 239: Is there a sample page to point to for a code example? Perhaps something that builds on top of the Navigation API and shows how the 2 integrate nicely.

explainer.md
```diff
@@ -1,751 +1,423 @@
...
 
-- `::page-transition-incoming-image` animates _from_ opacity 0.
-- `::page-transition-outgoing-image` animates _to_ opacity 0.
-- If an element exists in both Page-A and Page-B, a transform animation takes
-  its container from Page-A size/transform to Page-B size/transform. Otherwise
-  the element will remain in its final position.
+Since this feature is built on existing concepts such as pseudo-elements and CSS animations, tooling for this feature should fit in with existing developer tooling.
 
-All animations use a fill-mode of `both`.
+In Chrome's experimental implementation, the pre-existing animation panel could be used to debug transitions, and the pseudo-elements were exposed in the elements panel.
```
khushalsagar, line 270:
```suggestion
In Chrome's experimental implementation, the pre-existing animation panel could be used to debug transitions, and the pseudo-elements are exposed in the elements panel.
```

explainer.md
```diff
@@ -1,751 +1,423 @@
...
-- Open question: Are the generated keyframes usable outside the UA stylesheet?
-  If so, what are they called?
-- [Open question](https://github.com/WICG/shared-element-transitions/issues/84):
-  Default animations work well for things which are at least partially
-  in-viewport in both Page-A and Page-B, but it gets tricky if you consider a
-  non-sticky header that scrolled out of view by 1000s of pixels.
+# Future work
 
-All of these can be overridden by the developer via CSS animations.
+There are elements to this feature that we're actively thinking about, but aren't fully designed.
```
khushalsagar, line 276:
```suggestion
There are parts to this feature that we're actively thinking about, but aren't fully designed.
```

elements is too overloaded at this point. :P

explainer.md
```diff
@@ -1,751 +1,423 @@
...
-document.addEventListener("beforepageshow", async (event) => {
-  if (!event.transition) return;
-  document.querySelector(".header").style.pageTransitionTag = "header";
+const transition = new SameDocumentTransition();
+await transition.prepare(async () => {
+  await updateTheDOMSomehow();
+});
+```
+
+`prepare()` is used to signal that the current state should be captured, and the promise returned by the promise signals when the new state can be captured, and the transition can begin. Making cross-document transitions work means replacing those moments with moments either side of a cross-document navigation.
```
khushalsagar, line 291:
```suggestion
`prepare()` is used to signal that the current state should be captured, and the promise returned by the callback passed to the API signals when the new state can be captured, and the transition can begin. Making cross-document transitions work means replacing those events with implicit events on either side of a cross-document navigation.
```

explainer.md
```diff
@@ -1,751 +1,423 @@
...
+   └─ ::page-transition-container(child-item)
+      └─ ::page-transition-image-wrapper(child-item)
+         └─ …
 ```
-Note that the `page-transition-tag` CSS property is consulted at the time
-`start` is called, as well as after the passed callback is finished running to
-determine the source and destination elements respectively.
 
-# Relation to `element()`
+## Better pseudo-element selectors
```
khushalsagar, line 369: Could we add the content vs full capture point also above this? It has come up as a common use-case when the shape of a UI widget is changing in a transition.

explainer.md
```diff
@@ -1,751 +1,423 @@
...
-  re-fetched in the context of the new page to account for differences in
-  sandboxing.
+- Script can never read pixel content in the images. This is necessary since the document may embed cross-origin content (iframes, CORS resources, etc.) and multiple restricted user information (visited links history, dictionary used for spell check, etc.)
+- If an element is captured as a 'computed style + content image', any external resources specified on the container, such as background images, will be re-fetched in the context of the new page to account for differences in sandboxing.
 
-Cross-origin transitions aren't yet defined, but are likely to be heavily
-restricted.
+Cross-origin transitions aren't yet defined, but are likely to be heavily restricted.
 
 # Interactivity and accessibility
```
khushalsagar, line 417: Not sure if we should add this now since the details need to be sorted in the spec but one more relevant point:

* The duration for which DOM rendering is suppressed, to allow an author to asynchronously switch to the new DOM, input processing is also paused. This is necessary since the visual state presented to the user is inconsistent with the DOM state used for hit-testing.

explainer.md
```diff
@@ -1,751 +1,423 @@
...
 
-Page transitions are a purely visual affordance. In terms of interactivity,
-transition elements will behave like `div`s regardless of the original element.
-
-Developers could break this intent by adding interactivity directly to the
-transition element, e.g. by deliberately adding a `tabindex` attribute. But this
-isn't recommended.
+Page transitions are a purely visual affordance. In terms of interactivity, transition elements will behave like `div`s regardless of the original element.
```
khushalsagar, line 419: nit: Could these 3 be written as points.

explainer.md
```diff
@@ -1,751 +1,423 @@
...
-from the developer in future iterations. See issues
-[72](https://github.com/WICG/shared-element-transitions/issues/72) and
-[73](https://github.com/WICG/shared-element-transitions/issues/73) for detailed
-discussion on this topic.
-
-The single image mode works great for the share button and the root, as their
-transitions can be represented by simple transforms. However, the header changes
-size without stretching its shadow, and the content of the header moves
-independently and doesn't stretch. There's another mode for that:
+Page transitions not only look great, they also communicate direction of flow, and make it clear which elements are related from page to page. They can even happen during data fetching, leading to a faster perception of performance.
```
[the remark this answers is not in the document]
jakearchibald: I removed it because the "flash of white" is wrong, and it seemed to focus on MPA. But yeah, cutting it has lost too much. I'll add some of this back.

explainer.md
jakearchibald: I'll add a note

explainer.md
jakearchibald: Agreed. I avoided it because I wasn't sure if it'd be clear the styles were generated rather than static.

explainer.md
jakearchibald: Oh, so it starts with the incoming size, and animates from the outgoing size?

explainer.md
jakearchibald: We can chat about this later, as it's not 100% clear to me

explainer.md
jakearchibald: Wellllll they don't really integrate nicely right now, as you get that "did the navigation work or not??" issue that we're going to fix in the next iteration of the API.

explainer.md
jakearchibald: Hmm, but "Developers could break this intent" seems like a follow-on statement, where as a bulleted list is usually for independent statements.

jakearchibald:
Merging this just in case TAG take a look early. Other issues can be addressed in a follow-up commit after our meeting.

explainer.md
khushalsagar: Yup. The width/height corresponds to the to state and the keyframe provides the from state. Internally its the outgoing size until we hit the running phase but since this is web exposed when we get to the running/animation phase, at that point all values are from the incoming one.

explainer.md
khushalsagar: Fair. So we can update this once that issue is fixed in the spec.

explainer.md
khushalsagar: Sent another PR for this.

explainer.md
khushalsagar: I missed that. So 2 points. :) I put this in a follow up PR.

PR #194, opened 2022-09-20
Add md file for rough scoped transitions design

khushalsagar:


scoped-transitions.md
```diff
@@ -0,0 +1,110 @@
...
+   │  └─ ::page-transition-image-wrapper(outer)
+   │     ├─ ::page-transition-outgoing-image(outer)
+   │     └─ ::page-transition-incoming-image(outer)
+   └─ ::page-transition-container(inner)
+      └─ ::page-transition-image-wrapper(inner)
+         ├─ ::page-transition-outgoing-image(inner)
+         └─ ::page-transition-incoming-image(inner)
+
+Note: We will probably need to change the naming, as 'page-transition' doesn't work.
+Note: We might want to add an ID for a transition to control which transition the `page-transition-tag` applies to.
```
jakearchibald, line 49: We don't do this for root transitions. Feels like we should do it for both if we think it's worth it.

scoped-transitions.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+The algorithm for executing the transition is as follows:
+
+1. At the next rendering opportunity after `createTransition`, the browser saves the painted output and geometry information for each tagged element under scoped-transition-root. The details are described in [7.3.4](https://drafts.csswg.org/css-shared-element-transitions-1/#perform-an-outgoing-capture-algorithm) except transforms are computed relative to the element. An async task is used to dispatch the `updateDOM` callback after this step.
```
jakearchibald, line 54:
```suggestion
1. At the next rendering opportunity after `createTransition`, the browser saves the painted output and geometry information for each tagged element under scoped-transition-root. The details are described in [7.3.4](https://drafts.csswg.org/css-shared-element-transitions-1/#perform-an-outgoing-capture-algorithm) except transforms are computed relative to the element. A task is queued to dispatch the `updateDOM` callback after this step.
```

scoped-transitions.md
```diff
@@ -0,0 +1,110 @@
...
+
+2. For every subsequent rendering opportunity, the browser renders the sub-tree under scoped-transition-root using the cached output captured in step 1. This is done by generating the pseudo-element sub-tree (originating from scoped-transition-root) as described in [7.9](https://drafts.csswg.org/css-shared-element-transitions-1/#create-transition-pseudo-elements-algorithm) except only outgoing-image pseudo-elements are generated. This allows the developer to update scoped-transition-root's subtree without presenting it to the user.
+
+   Note: Pseudo-elements at this stage don't need to be developer exposed so UAs can use an alternate approach for this step. Using pseudo-elements (and default browser CSS) here helps define a consistent behavior for snapshots are updated if the scoped-transition-root is resized.
```
jakearchibald, line 58: I assume a spec-compliant implementation would need to match the pseudo-element rendering? I'm thinking about cases where the pseudo-element rendering differs from the real rendering due to layering.

scoped-transitions.md
```diff
@@ -0,0 +1,110 @@
...
+3. The captured output above is painted in a stacking context which is a descendant of scoped-transition-root's stacking context. The above example generates the following stacking context hierarchy.
+
+   ```
+   root
+   └─ outer
+      ├─ inner
+      └─ outer-page-transition (pseudo-tree underneath)
+   ```
+
+   If the scoped-transition-root is tagged, it's snapshot uses [content-capture](https://github.com/WICG/shared-element-transitions/blob/main/explainer.md#more-granular-style-capture) mode. This means that only the element's descendants (including both text and elements) are painted in the snapshot while box decorations (like border, box-shadow) continue to paint in the outer stacking context.
```
jakearchibald, line 69: I had the same thought, but worried it was weirdly different to what happens in root transitions. Also, are you proposing the full "granular style capture" model here? As in, what would happen if outer's background went from red to green. Would it animate (as in "granular style capture"), or switch (as in, we're just capturing the contents as a texture)?

scoped-transitions.md
```diff
@@ -0,0 +1,110 @@
...
+     <div class="inner"></div>
+   </div>
+   ```
+
+   ```js
+   document.createTransition(…);
+   document.querySelector('.outer').createTransition(…);
+   ```
+
+   There will be an image pseudo-element for `inner` in transitions on document and `outer` and its ambiguous which pseudo-element it should be painted in.
```
jakearchibald, line 98:
```suggestion
   Both transitions want to capture `inner` and `outer`. It's ambiguous which pseudo-elements they should be painted in.
```

Because of this, we should provide an API to get the active transitions an element is involved in, so it can choose to wait for them.

scoped-transitions.md
```diff
@@ -0,0 +1,110 @@
...
+
+2. The element which is the root for a transition must have `contain: layout`. This ensures it generates a stacking context and its painted output can be captured as an atomic unit.
+
+## Selector Syntax
+
+The CSS selector options for the pseudo-trees can be:
+
+- [part](https://w3c.github.io/csswg-drafts/css-shadow-parts-1/#part) selector which is spec'd to match elements in shadow DOM originating from an element. `html::part(container)` for document transitions and `#outer::part(container)` if outer is the scoped-transition-root. This implies that `#outer` can not be a shadow host for author's shadow DOM since which shadow DOM `#outer::part(container)` is targeting is ambiguous.
```
jakearchibald, line 106: Yeah this seems bad 😄

scoped-transitions.md
```diff
@@ -0,0 +1,110 @@
...
+
+- Pseudo-element syntax such as `html::page-transition::container(foo)` and `#outer::page-transition::container(foo)`.
+
+- `transition(…)` which is a new CSS function. The syntax is similar to `::part` but is limiting to page-transition elements originating from a DOM element. For example, `html::transition(container)` and `#outer::transition(container)`. This selector could be backed by a shadow DOM or pseudo-element based implementation.
```
jakearchibald, line 110: > This selector could be backed by a shadow DOM or pseudo-element based implementation.

The same could be said for the pseudo-element syntax.

scoped-transitions.md
```diff
@@ -0,0 +1,110 @@
...
+   ```js
+   document.createTransition(…);
+   document.querySelector('.outer').createTransition(…);
+   ```
+
+   There will be an image pseudo-element for `inner` in transitions on document and `outer` and its ambiguous which pseudo-element it should be painted in.
+
+2. The element which is the root for a transition must have `contain: layout`. This ensures it generates a stacking context and its painted output can be captured as an atomic unit.
+
+## Selector Syntax
```
jakearchibald, line 102: I think we can drop the section and save it for a shadow-dom-or-not comparison.

scoped-transitions.md
```diff
@@ -0,0 +1,110 @@
...
+
+```js
+element.createTransition({
+  updateDOM() {
+    /* … */
+  },
+});
+```
+
+The element becomes the **scoped-transition-root** for the transition. It's the element that will host the pseudo-element tree and page-transition-container sub-trees will be created for descendants that have a `page-transition-tag`. For example, the following page:
```
vmpstr, line 13: I'd add a disclaimer right at the top saying that this document talks about the pseudo-element tree as the implementation of the scoped transitions. Some considerations for a different implementation, Shadow DOM, are considered at the end

scoped-transitions.md
vmpstr: I think this is more a multiple independent transitions problem, but it's definitely something we need to figure out. 

Specifically, there is a problem with nested transitions where we have an element transition and, say, a document transition. Currently there is no way to distinguish what `page-transition-tag` means: should we tag that in the element transition or in the root transition or both? I don't think it's as easy as saying `::page-transition(transition-tag)`. 

It's possible that the page-transition-tag syntax should change to something like `page-transition-tag: (tag, [transition-tag1, transition-tag2])` or some tuple to indicate that this element has a tag that references one or more `transition-tag`s

scoped-transitions.md
```diff
@@ -0,0 +1,110 @@
...
+
+4. Once the `updateDOM` callback is finished, incoming-image pseudo-elements are added to the tree generated in step 2. Default animations are added as defined in [7.8](https://drafts.csswg.org/css-shared-element-transitions-1/#animate-a-page-transition-algorithm).
+
+## Constraints
+
+An element which is the root for a transition has the following constraints. These constraints may stop being satisfied at any phase of the transition lifecycle and we'll need to define the fallback behavior for such cases. The simplest option is to consider them a developer error and [skip the transition](https://drafts.csswg.org/css-shared-element-transitions-1/#skip-the-page-transition).
+
+1. There shouldn't be an active transition on the element, its descendants or its ancestors. This is because a transition involves taking the painted output of an element and rendering it into a pseudo-element. For example, in the following case:
```
vmpstr, line 77: I think this is a somewhat severe limitation, something like a component having a transition would then preclude the embedding page from doing a transition and vice versa

scoped-transitions.md
vmpstr: True. I feel like the function-like syntax yields itself to simpler and smaller selectors though, so maybe we should consider it regardless of the structure backing

scoped-transitions.md
vmpstr, line 54: Is this the part that also describes render blocking? If we need an element level render blocking, I think we need to mention this explicitly, since it's one of the more complicated parts of the proposal

scoped-transitions.md
khushalsagar: The idea comes from the comment you made. :) I thought a lot about it and realized it makes sense. If we try to capture the root snapshot (similar to root transitions) then it becomes difficult to describe the behaviour of clips + effects.

For example, scoped-transition-root has overflow:clip and a blur. The blur will draw outside the box because overflow:clip only affects the content (not self decorations). But the pseudo-element displaying scoped-transition-root's image will be its descendant. So it should clip the image and as a result will clip these box decorations.

> are you proposing the full "granular style capture" model here? As in, what would happen if outer's background went from red to green. Would it animate (as in "granular style capture"), or switch (as in, we're just capturing the contents as a texture)?

The latter, it'll be a direct flip. Its simpler because if the box resizes we can't avoid doing a relayout and repaint. And we'll have to teach the style/layout stack to use the old style values when painting these effects. If it's a direct flip, then style/layout/paint can proceed as usual. And way later we just skip drawing the tree under scoped-transition-root since the snapshot is drawing on top of it (same approach as today). Curious to hear @vmpstr's thoughts on this too.

scoped-transitions.md
khushalsagar: I figured we can finish the transition on all descendants if an ancestor tries to do one. But if we wanted to avoid this then at the very least we need this: "No descendant of scoped-transition-root should be a shared element in any other transition".

If the scoped-transition-root itself is tagged, I could see us taking its entire painting (with its pseudo-tree) and placing it in a image pseudo-element participating in a document transition. But if one of its descendants is tagged in a root transition, then its unclear which pseudo-element its painting should go to.

I think the above is reasonable since we're talking about independent components driving transitions so the document ideally shouldn't tag elements within a transitioning widget anyway?

scoped-transitions.md
khushalsagar: +1

scoped-transitions.md
khushalsagar: Sounds good. I'll remove it. Is it worth keeping the point about why ::part is bad because of this feature though? The goal behind doing this design now was so we can point to it to veto the ::part syntax.

scoped-transitions.md
khushalsagar: > I feel like the function-like syntax yields itself to simpler and smaller selectors

+1. This was why I initially went with html::page-transition-container() too.

scoped-transitions.md
khushalsagar: Done.

scoped-transitions.md
jakearchibald: In a DOM structure like:

```html
<root>
  <component-a>
    <component-b></component-b>
  </component-a>
</root>
```

Why are we creating special syntax to differentiate a root transition and a scoped transition, but not worried about differentiating between transitions scoped to component-a vs component-b?

khushalsagar: > but not worried about differentiating between transitions scoped to component-a vs component-b

Differentiating between these is also a goal. For the example above:

```html
<root>
  <component-a>
    <div class="header-a"></div>
    <component-b>
      <div class="header-b"></div>
    </component-b>
  </component-a>
</root>
```

The syntax could be:
```css
.component-a {
  page-transition-root: a;
}
.component-b {
  page-transition-root: b;
}

.header-a {
  page-transition-tag: header a;
}

.header-b {
  page-transition-tag: header b;
}
```

Devs could always create prefixes themselves if they wanted this, `page-transition: header-a`, so its just syntactic sugar. We can ignore it for now.

khushalsagar: Moved to issue #195.

scoped-transitions.md
khushalsagar: Added a section to the doc to capture this.

scoped-transitions.md
khushalsagar: Left a note in the doc.

scoped-transitions.md
khushalsagar: I left a comment on your doc to cover this there.

scoped-transitions.md
khushalsagar: Let's have this discussion in the doc focused on the CSS/JS API.

scoped-transitions.md
khushalsagar: Added a section in the doc to cover this.

scoped-transitions.md
khushalsagar: The "Suppressing rendering" now pulls this discussion out separately too.

MikhaelTovel (approved):

scoped-transitions.md
```diff
@@ -0,0 +1,132 @@
+# Scoped Transitions
+
+This doc is a summary of early design explorations to scope transitions to a DOM sub-tree. It uses the pseudo-element tree to describe the UA generated tree to refer to the existing spec. But considerations for an alternate Shadow DOM implementation are covered in *TODO: add link to jake's doc*.
```
jakearchibald, line 3:
```suggestion
This doc is a summary of early design explorations to scope transitions to a DOM sub-tree. It uses the pseudo-element tree to describe the UA generated tree to refer to the existing spec. A [separate doc considers an alternate Shadow DOM implementation](https://docs.google.com/document/d/1kW4maYe-Zqi8MIkuzvXraIkfx3XF-9hkKDXYWoxzQFA/edit?usp=sharing).
```

jakearchibald (approved):

PR #198, opened 2022-11-07
Renames for SPA API and add tentative MPA API

khushalsagar:


explainer.md
```diff
@@ -345,6 +341,114 @@ And here's the result:
 
 https://user-images.githubusercontent.com/93594/184120371-678f58b3-d1f9-465b-978f-ee5eab73d120.mp4
 
+## Cross-document same-origin transitions
+
+This section outlines the navigation specific aspects of the ViewTransition API. The rendering model for generating snapshots and dispalying them using a tree of targetable pseudo-elements is the same for both SPA/MPA.
```
jakearchibald, line 346:
```suggestion
This section outlines the navigation specific aspects of the ViewTransition API. The rendering model for generating snapshots and displaying them using a tree of targetable pseudo-elements is the same for both SPA/MPA.
```

explainer.md
```diff
@@ -345,6 +341,114 @@ And here's the result:
...
+
+### Declarative opt-in to transitions
+
+The first step is to add a new meta tag to the old and new Documents. This tag indicates that the author wants to enable transitions for same-origin navigations to/from this Document.
+
+```html
+<meta name="view-transition" content="same-origin">
```
jakearchibald, line 353: Is there a technical reason we couldn't have this signal in the CSS? It feels like it would benefit from being wrapped in media queries, particularly `prefers-reduced-motion`.

explainer.md
```diff
@@ -345,6 +341,114 @@ And here's the result:
...
+
+Issue: The meta tag can be used to opt-in to other navigation types going forward: same-document, same-site, etc.
+
+### Script Events
+
+Script can be used to customize a transition based on the URL of the old/new Document; or the current state of the Document when the transition is initiated. The Document could've been updated since first old from user interaction.
+
+#### Script on Old Document
+```js
+document.addEventListener("viewTransitionOnOldDocument", (event) => {
```
jakearchibald, line 378: Bikeshedding comes later I guess, but modern DOM event names are lowercasenospaces.

explainer.md
```diff
@@ -345,6 +341,114 @@ And here's the result:
...
+  // Cancel the transition (based on new URL) if needed.
+  if (shouldNotTransition(event.url)) {
```
jakearchibald, line 380: Is `event.url` after redirects? I wonder if something in the navigation API is a good place for this.

explainer.md
```diff
@@ -345,6 +341,114 @@ And here's the result:
...
+  // start. This can be customized based on the old Document state when the
+  // transition was initiated.
+  markRenderBlockingResources(oldNames, context);
+  
+  // The browser defers painting the new Document until all render-blocked
+  // resources have been fetched or timed out.
+  // The `domUpdated` promise can be used to inspect when the above is
+  // done and add viewTransitionNames based on what the new Document's
+  // first paint would look like.
+  transition.domUpdated.then(() => {
```
jakearchibald, line 424: I don't think we should reuse `domUpdated` for this, as it doesn't really mean the same thing. We could add a new promise for it if needed. `transition` doesn't need to be the exact same type.

jakearchibald (commented):
I like where this is heading. I'll have a think around the API tomorrow too and write some more detailed notes.

explainer.md
```diff
@@ -345,6 +341,114 @@ And here's the result:
...
+  transition.finished.then(() => {
+    // Remove viewTransitionNames tied to this transition.
+    thumbnail.style.viewTransitionName = "none";
+  });
+});
+```
+
+This provides the same scripting points as the SPA API, allowing developers to set class names to tailor the animation to a particular type of navigation.
+
+Issue: Do we need better timing for viewTransitionOnNewDocument event? Especially for Documents restored from BFCache.
```
jakearchibald, line 449: Fwiw, I like `beforepagehide` (or maybe just `pagehide`??) and `beforepageshow`.

explainer.md
khushalsagar: The prefers-reduced-motion query is a good point. We can make this work in CSS.

Where should this CSS property be though? It's sort of a global knob so doesn't make sense for it to be a per-element property that participates in the style cascade.

explainer.md
khushalsagar: Yes. The implementation I have waits to have all redirects resolve and get a response from the server for the new Document before dispatching this event (and initiating snapshoting). There is a trade-off here:

- Snapshot too early and the current Document's rendering freezes/it becomes inert too soon. We can't keep rendering frames for it for the same rendering suppression problem that SPA has.
- Snapshot too late and we won't parallelize all the snapshoting work with the network fetch of the new page.

We'll get a better idea of optimizing this with some demos etc.

explainer.md
khushalsagar: I can see the benefit of pulling domUpdated into an independent API for render-blocking in general. But other than that, do the ready/finished promise make sense? They are conceptually the same and needed for the same use-case as the SPA API. I can move them into a different ViewTransition type.

explainer.md
khushalsagar: Do we need a generic event for this? I was hoping for something ViewTransition specific so we can skip dispatching if there is no developer opt-in to the API. Don't want to add more events that must be dispatched for all navigations (especially the back/fwd ones which can be completely browser driven).

We'll also not dispatch these events if the browser decided to skip the transition for any reason.

jakearchibald: Yeah, I'm coming round to the specific event idea.

Can't help myself bikeshedding:

- `documenttransitionoldcapture` - fired on the outgoing page. Can't think of anything shorter.
- `documenttransition` - fired on the new page.

It feels like "document transition" is right here, to differentiate it from SPA transitions.

jakearchibald:
One thing to raise with security: This creates a brief moment on the current page where it can know:

- The navigation started with a request to a same-origin page (and the specific URL)
- The navigation ended on a different-origin page (but no specific URL)

You can figure this out in a number of different ways, with iframes and session history, but knowing on the current page before the navigation commits is new I think.

explainer.md
khushalsagar: Discussed offline. We'll file an issue to bring this up with CSSWG. The media query for prefers-reduced-motion is an example for why it's better for it to be in CSS but we need to establish a pattern for global CSS settings.

explainer.md
khushalsagar: Given the discussion below for a feature specific event here, can we resolve this?

explainer.md
khushalsagar: Removed domUpdated promise with an open issue for providing developers with a signal to detect when DOM is ready for first paint. So they could conditionally decide what is tagged.

khushalsagar:
Since we dispatch the event only after all redirects are resolved, there shouldn't be a case where we switch from same-origin to cross-origin final URL after the event is dispatched.

khushalsagar:
Oh yeah, didn't think it that way. I don't expect it to be an issue since as you mentioned, this can already be known after the navigation in multiple ways. But added a note to highlight this in the security review.

khushalsagar:
One thought about name, I've been trying to avoid the word "cross-document" since we'd likely also do this for same-document navigations. It's fine for now though.

PR #208, opened 2023-05-17
Add explainer for cross-document navigations

noamr:


cross-doc-explainer.md
```diff
@@ -0,0 +1,114 @@
...
+To enable cross-document transitions, both the old and new document need to add a meta tag. This tag
+indicates that the author wants to enable transitions for same-origin navigations to/from this
+Document.
+
+
+### Issues
+
+1. The meta tag can be used to opt-in to other navigation types going forward: same-document, same-site, etc.
+
+1. Using meta-tags prevents the declaration being controlled by media queries, which feels important for `prefers-reduced-motion`.
```
vmpstr, line 56: There's a discussion about having opt-in be in css as well: https://github.com/w3c/csswg-drafts/issues/8048

cross-doc-explainer.md
```diff
@@ -0,0 +1,114 @@
...
+Cross-document transitions should work by default without JavaScript intervention, and should have
+the right JavaScript knobs for when the defaults are not enough.
+
+# How it works
+
+## In a nutshell
+Documents declare that they support same-origin cross-document transitions, using a
+[meta tag](#opt-in). When a navigation between two such documents takes place (and has no
+cross-origin redirects in the middle), the state of the old document is captured, in the same
+manner as same-origin transitions. When the new document is ready for the transition (i.e., when all render-blocking resources are ready), the new state is captured,
```
vmpstr, line 32: Just a nit about the phrasing "the new state is captured" which makes it sounds like it's a one-time snapshot. I think a more pedantic statement is that the new shared elements are discovered and their state is tracked for the duration of the animation

cross-doc-explainer.md
```diff
@@ -0,0 +1,114 @@
...
+## Lifecycle
+
+![Lifecycle chart for cross-document transitions](media/mpa-chart.svg)
+
+### Capturing the old state
+
+The old state is captured right before the old document is hidden and a new one is shown.
+In the HTML spec this is defined [here](https://html.spec.whatwg.org/#populating-a-session-history-entry:loading-a-document).
+This can either happen during normal navigations, when the new document is about to be created,
+in BFCache navigations, or when activating a prerendered document.
```
vmpstr, line 67:
```suggestion
in Back/Forward cache navigations, or when activating a prerendered document.
```

cross-doc-explainer.md
```diff
@@ -0,0 +1,114 @@
...
+1. Using meta-tags prevents the declaration being controlled by media queries, which feels important for `prefers-reduced-motion`.
+
+## Lifecycle
+
+![Lifecycle chart for cross-document transitions](media/mpa-chart.svg)
+
+### Capturing the old state
+
+The old state is captured right before the old document is hidden and a new one is shown.
+In the HTML spec this is defined [here](https://html.spec.whatwg.org/#populating-a-session-history-entry:loading-a-document).
```
vmpstr, line 65: "this" seems ambiguous, it's not the capturing that's defined. To clarify my understanding, do you mean that the document hiding/showing is defined here, so the capture steps would have to reference this section?

cross-doc-explainer.md
```diff
@@ -0,0 +1,114 @@
...
+* Abort the transition if certain conditions apply.
+
+* Modify the pseudo-element style based on certain conditions, e.g. apply transition
+   names only to images that are already loaded.
+
+* Prepare the exit transition right before capture.
+
+There are several options as to how to enable this:
+
+1. Fire an event right before the new document capture. This would be either right
```
vmpstr, line 98: Similar comment to above, "new document capture" sounds like a misnomer. Maybe "new document's shared element discovery" or something?

cross-doc-explainer.md
```diff
@@ -0,0 +1,114 @@
...
+   before the first `requestAnimationFrame` callback, or after `pageshow` in the reactivation case. Potential name: `reveal`. Note that in the latter case, it's too late to delay the transition via the render-blocking mechanism. This event doesn't
```
vmpstr, line 99: "reveal" and "pageshow" sound very similar in naming tbh, which makes me think that pageshow might be the right event? maybe?

noamr: Yea also `visibilitychange`. The problem with `pageshow` is that in the case of a new document, it will wait until the `load` event. So we could use it for reactivate and use the first rAF callback for new document, but that means that developers would have to be extra careful and always use those two callbacks. Another option is that we have a `document.currentTransition` or some such, and `requestAnimationFrame` is *always* the place to handle it.

cross-doc-explainer.md
noamr: Isn't it [capture the new state](https://drafts.csswg.org/css-view-transitions-1/#capture-new-state-algorithm) the SPA algorithm though?

cross-doc-explainer.md
noamr: Will reword to "that moment"

cross-doc-explainer.md
vmpstr: I see. Yeah that's why I think the diagram you have is useful: the relative timing of everything here is hard

I'm personally not a fan of document.currentTransition, since it feels like we still need to define when that object is set and what is its lifetime. For example, in prerender it wouldn't exist always, so at some point we'd populate the currentTransition. It feels like the time at which we would populate it is also the time where we could fire an event. This isn't an opinion that I hold strongly though

cross-doc-explainer.md
noamr: Added a link to that discussion

cross-doc-explainer.md
noamr: I made these references links to the "capture the old/new state" algorithms. If we want to rename them we can but I don't want to do that in the explainer if we don't change the spec.

IMO capture is OK though

cross-doc-explainer.md
noamr: See comment above, remaining consistent with the spec

cross-doc-explainer.md
```diff
@@ -0,0 +1,114 @@
...
+
+Before creating the new document (or activating a cached/prerendered one), the UA would update the rendering and snapshot the old document, in the same manner a document is snapshotted for a same-document navigation.
+
+Note that currently there are no planned new events for the exit transition.
+The developer can use existing events like `navigate` or `click` to customize the
+exit transition at moments that should be late enough.
+
+#### Issues
+
+1. Should we enable skipping the transition on exit? This could be done on`pagehide`.
```
khushalsagar, line 78: Is `pagehide` guaranteed to be before the new Document is created? It would be awkward for the old Document to skip the transition in this event if the new Document has already accessed it via document.activeTransition?

cross-doc-explainer.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Capturing the new state
+
+The [new state is captured](https://drafts.csswg.org/css-view-transitions-1/#capture-new-state-algorithm) at the first [rendering opportunity](https://html.spec.whatwg.org/#rendering-opportunity)
```
khushalsagar, line 82: We need to discuss whether this can happen during the rendering opportunity. Let's take this up on a csswg issue and link it here?

cross-doc-explainer.md
```diff
@@ -0,0 +1,114 @@
...
+
+For most cases, opting in via the meta tag, styling with the pseudo-elements, and delaying the transition using the
+[render-blocking mechanism](https://html.spec.whatwg.org/#render-blocking-mechanism) should be sufficient. But there are certain cases where further customization is desired, for example:
+
+* Abort the transition if certain conditions apply.
+
+* Modify the pseudo-element style based on certain conditions, e.g. apply transition
+   names only to images that are already loaded.
+
+* Prepare the exit transition right before capture.
```
khushalsagar, line 95: nit: Didn't we already cover this under "Capturing the old state"?

cross-doc-explainer.md
```diff
@@ -0,0 +1,114 @@
...
+
+There are several options as to how to enable this:
+
+1. Fire an event right before the new document capture. This would be either right
+   before the first `requestAnimationFrame` callback, or after `pageshow` in the reactivation case. Potential names: `reveal`, `beforepagetransition`. Note that in the latter case, it's too late to delay the transition via the render-blocking mechanism. This event doesn't have to be specific to transitions.
```
khushalsagar, line 100: A note about why we can't use `pageshow` is worth adding. Seems like it should work for the restore cases but its too late for the new load case.

cross-doc-explainer.md
```diff
@@ -0,0 +1,114 @@
...
+
+1. Expose something like a `document.currentRevealTransition` or so that the script can
```
khushalsagar, line 102: I don't think its about choosing between one of these options, you need both. First one for customizing the animation based on what was fetched before first paint. And second one for customizing what is render blocking.

cross-doc-explainer.md
```diff
@@ -0,0 +1,114 @@
...
+   query or skip. We might also want to enable delaying the transition with a promise,
```
khushalsagar, line 103: > We might also want to enable delaying the transition with a promise

Big +1 to this idea but as a generic "delay first paint" concept, for both new Documents and restore from BFCache/pre-render.

cross-doc-explainer.md
```diff
@@ -0,0 +1,114 @@
...
+   the same way that's available for same-document transitions.
+
+### Issues
+
+1. Customizing which resources are render-blocking in `reveal` requires it to be dispatched before parsing `body`, or explicitly allow render-blocking resources to be added until this event is dispatched.
+
+1. We'd likely need an API for the developer to control how much Document needs to be fetched/parsed before the transition starts, or delay the capture with a promise.
```
khushalsagar, line 110: The "delay transition with a promise" idea helps with that.

cross-doc-explainer.md
```diff
@@ -0,0 +1,114 @@
...
+
+1. The browser defers painting the new Document until all render-blocked resources have been fetched or timed out. Do we need an explicit hook for when this is done or could the developer rely on existing `load` events to detect this? This would allow authors to add viewTransitionNames based on what the new Document's first paint would look like.
```
khushalsagar, line 112: `load` is fired too late, once all resources (including iframes) have been fetched. We don't have any event which is fired when just the render-blocking resources are fetched. First rAF is the closest proxy to that.

cross-doc-explainer.md
```diff
@@ -0,0 +1,114 @@
...
+
+1. Should we allow access to the transition once its started? Perhaps access to a [ViewTransition](https://drafts.csswg.org/css-view-transitions-1/#viewtransition) object or similar?.
```
khushalsagar, line 114: Yes please. The `finished` promise and `skipTransition` API are useful even after the transition has started.

explainer.md
```diff
@@ -54,7 +53,7 @@ and [Windows](https://docs.microsoft.com/en-us/windows/apps/design/motion/page-t
 
 # MPA vs SPA solutions
```
khushalsagar, line 54: The information is out of date, its not experimental anymore and you don't need a flag.

cross-doc-explainer.md
noamr: I mean relying on `load` events of the blocking iframes/images

media/mpa-chart.svg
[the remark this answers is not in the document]
noamr: Yes.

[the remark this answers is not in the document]
noamr: Fixed

yoavweiss (commented):
A couple of typos

cross-doc-explainer.md
noamr: For now I'm saying "right before the first rendering opportunity".

cross-doc-explainer.md
noamr: Yes it's guaranteed!

cross-doc-explainer.md
```diff
@@ -0,0 +1,163 @@
...
+cross-document transitions as well. This document provides explanations about the additional
+semantics, and how cross-document transitions work.
+
+# Design Principles
+
+## Compatible with same-document transitions
+
+Developers shouldn't have to jump through hoops or rethink their transition design when switching
+between an MPA architecture and an SPA architecture. The main building blocks of the transition,
+the way the states are captured, the way the captured images are animated, and the JavaScript API,should remain the same, as much as possible.
```
khushalsagar, line 21:
```suggestion
the way the states are captured, the way the captured images are animated, and the JavaScript API, should remain the same, as much as possible.
```

cross-doc-explainer.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Declarative opt-in
+
+To enable cross-document transitions, the old and new documents need to be coordinated with each
+other - as in, the transition names in the old document match the ones in the new document and the
+effect of animating between them is intentional. Otherwise, there could be a situation where two
+documents of the same origin define styles for same-document transitions independently, and enabling
+this feature would create an unexpected transition between them.
+
+This is an issue that is specific to cross-document transitions, as single-document transitions are
```
khushalsagar, line 72:
```suggestion
This is an issue that is specific to cross-document transitions, as same-document transitions are
```

cross-doc-explainer.md
```diff
@@ -0,0 +1,163 @@
...
+of the new document. This allows the new document to use the
+[render-blocking mechanism](https://html.spec.whatwg.org/#render-blocking-mechanism) as a way to
+delay the transition.
+
+As shown in the chart above, that first rendering opportunity can come in two cases, either
+it's a newly initialized document that's no longer [render-blocked](https://html.spec.whatwg.org/multipage/dom.html#render-blocked), or it's a document that's been frozen due to back-forward cache
+or prerendered, and is now being activated.
+
+Note: relying on the [render-blocking](https://html.spec.whatwg.org/multipage/dom.html#render-blocked) mechanism is limited, as only a limited set of elements participate in that
+mechanism. See [proposal to extend it](whatwg/html#9332).
```
khushalsagar, line 134:
```suggestion
mechanism. See [proposal to extend it](https://github.com/whatwg/html/issues/9332).
```

cross-doc-explainer.md
```diff
@@ -0,0 +1,163 @@
...
+observed](https://drafts.csswg.org/css-view-transitions-1/#dom-viewtransition-finished).
+
+To accomplish the same control and observability, the developer would need access to a
+[ViewTransition](https://drafts.csswg.org/css-view-transitions-1/#the-domtransition-interface)
+object. There are several ways to approach this, and this is an open issue:
+
+1. Expose it via `document.pendingViewTransition`. This would be available only before the document
+gets render-unblocked for the first time, and right at reactivation.
+
+1. Fire a special `reveal` event at both lifecycle moments, with a `ViewTransition` object. See whatwg/html#9315, w3c/csswg-drafts#8682, and w3c/csswg-drafts#8805. Note that this event is
```
khushalsagar, line 151: Mind putting a directly clickable link for the html issues? Copying whatwg/html#9315 doesn't work.

cross-doc-explainer.md
```diff
@@ -0,0 +1,163 @@
...
+different from [`pageshow`](https://html.spec.whatwg.org/#event-pageshow) as in the newly
+initialized document `pageshow` is only fired once the document is fully loaded.
+
+1. Add an API that allows continuing a view transition, using the current state as the old state.
```
khushalsagar, line 155: This will be a completely new feature for same-doc transitions also (see the thread here : https://chat.google.com/room/AAAAaLJ4XZQ/HrGp3_YjvAw). Not sure if it makes sense to mention here since we're talking about what's needed in the script API for parity with same-doc?

khushalsagar (approved):

cross-doc-explainer.md
noamr: Is that necessary? Github links work magically in Github MarkDown

cross-doc-explainer.md
noamr: The idea is that if we have that feature, we might not need additional MPA-specific APIs.

cross-doc-explainer.md
noamr: Done

cross-doc-explainer.md
noamr: I reworded this a bit. How does it look now @khushalsagar?

PR #211, opened 2023-06-25
Update cross-doc-explainer.md

noamr:
Updated to be more precise about current thinking re. API shape.

### DIFF
--- a/cross-doc-explainer.md
+++ b/cross-doc-explainer.md
@@ -147,8 +147,7 @@ See [whatwg/html#9315](https://github.com/whatwg/html/issues/9315), w3c/csswg-dr
 Note that this event is different from [`pageshow`](https://html.spec.whatwg.org/#event-pageshow) as
 in the newly initialized document `pageshow` is only fired once the document is fully loaded.
 
-A potential alternative would be to expose it via `document.activeViewTransition`. This would be available only before the document
-gets render-unblocked for the first time.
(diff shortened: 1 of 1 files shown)


cross-doc-explainer.md
```diff
@@ -143,15 +142,14 @@ observed](https://drafts.csswg.org/css-view-transitions-1/#dom-viewtransition-fi
...
-gets render-unblocked for the first time, and right at reactivation.
-
-1. Fire a special `reveal` event at both lifecycle moments, with a `ViewTransition` object. See [whatwg/html#9315](https://github.com/whatwg/html/issues/9315), w3c/csswg-drafts#8682, and w3c/csswg-drafts#8805.
+object. To achieve that, we propose to fire a `reveal` event at both lifecycle moments, with a `ViewTransition` object.
+See [whatwg/html#9315](https://github.com/whatwg/html/issues/9315), w3c/csswg-drafts#8682, and w3c/csswg-drafts#8805.
 Note that this event is different from [`pageshow`](https://html.spec.whatwg.org/#event-pageshow) as
 in the newly initialized document `pageshow` is only fired once the document is fully loaded.
 
+A potential alternative would be to expose it via `document.activeViewTransition`. This would be available only before the document
+gets render-unblocked for the first time.
```
khushalsagar, line 151: > This would be available only before the document gets render-unblocked for the first time.

Are you trying to say that the accessor returns null once the Document is no longer render-blocked on initial load? It's obvious this misses the activation cases but they're technically doable. Would be better to have some text clarifying why we're not proposing this syntax. Something like: "figuring out when this should be queried for activation cases, in the absence of an event, is hard"?

khushalsagar (approved):

PR #214, opened 2023-08-24
Update document-render-blocking.md to include more options

khushalsagar:


document-render-blocking.md
```diff
@@ -20,14 +26,47 @@ The main use-case for this proposal is [cross-document View Transitions](https:/
 
 Identifying all animations correctly requires the browser to render-block the new Document until all elements which will be assigned a view-transition-name have been added to the DOM. Otherwise morph animations will turn into exit animations and entry animations will get skipped. Since browsers use heuristics to optimally yield and render when fetching a new Document, a consistent transition UX across all browsers is not feasible without an explicit hint from developers to delay rendering until the requisite nodes have been parsed.
 
-# Proposal
+# Other Use Cases
+
+There are a few other scenarios where a feature to control when the parser yields for rendering can be helpful:
+
+- Lower CLS: A stable layout of the DOM depends on both parsing the requisite DOM nodes and fetching relevant stylesheets. Without control over parsing, its likely that the browser does multiple renders with layout shifts as more of the DOM is parsed.
```
vmpstr, line 33:
```suggestion
- Lower CLS: A stable layout of the DOM depends on both parsing the requisite DOM nodes and fetching relevant stylesheets. Without control over parsing, its possible that the browser does multiple renders with layout shifts as more of the DOM is parsed.
```

document-render-blocking.md
```diff
@@ -20,14 +26,47 @@ The main use-case for this proposal is [cross-document View Transitions](https:/
...
+Authors will sometimes use `display: none` or `opacity: 0` to hide the whole Document to prevent this.
```
vmpstr, line 34:
```suggestion
Authors will sometimes initially set `display: none` or `opacity: 0` to hide the whole Document to prevent this, only showing it when the whole Document is parsed.
```

document-render-blocking.md
```diff
@@ -20,14 +26,47 @@ The main use-case for this proposal is [cross-document View Transitions](https:/
...
+
+- Atomic rendering of semantic elements: A UI widget built using a DOM sub-tree might not make sense to render partially. Consider a menu where only half the buttons show up on first render. Authors could mark sections in the Document which semantically render one widget so the browser doesn't yield midway through parsing one widget.
```
vmpstr, line 36: I feel like a parser may have some context here already, like try not yielding between siblings, or something like, but yeah your point that developers may want to control this makes sense

document-render-blocking.md
```diff
@@ -20,14 +26,47 @@ The main use-case for this proposal is [cross-document View Transitions](https:/
...
+
+- Optimal Yielding: The browser may yield later than was necessary leading to rendering more what's required for above the fold content. A developer hint about a yielding trigger could imply the first frame has less DOM to parse, style, and layout. Browsers can optimize paint to be limited to onscreen content but the prior stages are executed over the entire DOM.
```
vmpstr, line 38: I'm not sure I'm parsing the second and third sentences well.

Do you mean to say that a developer can provide a better hint resulting in less of the DOM tree to parse, style, and layout?

And also, "prior stages" you mean style & layout?

document-render-blocking.md
```diff
@@ -20,14 +26,47 @@ The main use-case for this proposal is [cross-document View Transitions](https:/
...
+However, its difficult for authors to know when the above-the-fold content ends given the varity of device form factors. It could also be served better by `content-visibility` which can optimize out style/layout for offscreen content.
```
vmpstr, line 39:
```suggestion
However, its difficult for authors to know when the above-the-fold content ends given the variety of device form factors. This situation could also be solved better by `content-visibility: auto` which can optimize out style/layout for offscreen content.
```

document-render-blocking.md
```diff
@@ -20,14 +26,47 @@ The main use-case for this proposal is [cross-document View Transitions](https:/
...
+
+These use-cases are not the primary problem targeted by this proposal, they are listed to evaluate whether the ability to block parsing should be limited to when the new Document will be displayed with a View Transition or all loads.
+
+# Dependencies
+
+The set of elements which need to block rendering for View Transition depends on which Document the user is coming from. A real world example is as follows.
+
+The user is navigating between Documents of a site which has a header. This header can be scrolled offscreen by the user, so its not guarenteed to be onscreen when a navigation is initiated. The following cases are possible:
```
vmpstr, line 47:
```suggestion
The user is navigating between Documents of a site which has a header. This header can be scrolled offscreen by the user, so it's not guaranteed to be onscreen when a navigation is initiated. The following cases are possible:
```

document-render-blocking.md
```diff
@@ -20,14 +26,47 @@ The main use-case for this proposal is [cross-document View Transitions](https:/
...
+
+## Blocking Attribute
 
 Add the [blocking attribute](https://html.spec.whatwg.org/#blocking-attributes) on the [HTMLHtmlElement](https://html.spec.whatwg.org/multipage/semantics.html#the-html-element). This is the [document element](https://dom.spec.whatwg.org/#document-element) for an HTML Document.
 
 The timing for when this attribute can block rendering for a Document is already well defined in [render-blocking mechanism](https://html.spec.whatwg.org/multipage/dom.html#render-blocking-mechanism). The user agent [unblocks rendering](https://html.spec.whatwg.org/multipage/dom.html#unblock-rendering) on this element when it's done parsing the document as defined [here](https://html.spec.whatwg.org/multipage/parsing.html#the-end).
 
-## Sample Code
-### Block Rendering on Full Document Parsing
+This approach neatly fits with the existing `blocking` primitive in html. The con is that while its trivial to block rendering until the full Document is parsed, more not-so-obvious code is needed to optimally block only on the minimal requisite set of elements. That makes it likely that authors will just block on full parsing since that will fix the correctness issues.
```
vmpstr, line 65: I would also mention that this would be bad, because although the correctness issue would be fixed, the overall user experience may degrade due to the length of time they need to wait before seeing the first frame

document-render-blocking.md
```diff
@@ -36,16 +75,104 @@ The timing for when this attribute can block rendering for a Document is already
...
+  blockingElementIds.delete("header-id");
+  maybeUnblockRendering();
+</script>
+</body>
+</html>
+```
+
+## Meta Tag with Blocking ElementIds
+
+Add a new meta tag with the name `blocking-elements` and `content` attribute set to the list of [element IDs](https://dom.spec.whatwg.org/#concept-id) which must be parsed before rendering. `*` is a special keyword which implies every ID should be blocking.
```
vmpstr, line 117:
```suggestion
Add a new meta tag with the name `blocking-elements` and `content` attribute set to the list of [element IDs](https://dom.spec.whatwg.org/#concept-id) which must be parsed before rendering. `*` is a special keyword which implies every element should be blocking.
```
The distinction is that if I don't specify an id on my elements, we should still block under `*` right?

vmpstr, line 117: Is this limited to parsing the shallow element, or does this include subresources? I'm thinking specifically if I have an img here, what are we blocked on exactly

document-render-blocking.md
```diff
@@ -36,16 +75,104 @@ The timing for when this attribute can block rendering for a Document is already
...
+
+Each Document has a render-blocking-until-parsed element ids set, initially empty. A Document is [render-blocked](https://html.spec.whatwg.org/#render-blocked) if this set is non-empty.
```
vmpstr, line 119: Or `*`? If it's a set of ids, and I have no ids, then maybe we need a special carve out for `*`

document-render-blocking.md
```diff
@@ -36,16 +75,104 @@ The timing for when this attribute can block rendering for a Document is already
...
+
+- If the value of the `content` attribute changes, and the Document [allows adding render blocking elements](https://html.spec.whatwg.org/#allows-adding-render-blocking-elements), the render-blocking-until-parsed element ids is set to the new attribute value.
+   This means authors can run script to configure the list until the body tag is parsed (after which no new render blocking resources can be added).
```
vmpstr, line 122:
```suggestion
   This means authors can run script to configure the list until the opening `<body>` tag is parsed (after which no new render blocking resources can be added).
```

document-render-blocking.md
```diff
@@ -36,16 +75,104 @@ The timing for when this attribute can block rendering for a Document is already
...
+
+- If the value of the `content` attribute changes, and the Document **does not** [allow adding render blocking elements](https://html.spec.whatwg.org/#allows-adding-render-blocking-elements), render-blocking-until-parsed element ids is set to be an intersection of the existing value and the new attribute value.
+   This means authors can run script to remove render-blocking elements after the body tag is parsed but can't add more elements. This allows authors to implement their own timeout if needed.
+
+- Each time an element's ID value changes, the browser checks if the set of elements which have been completely parsed (i.e. the end tag has been parsed) include all IDs in the render-blocking-until-parsed element ids set. If yes, the render-blocking-until-parsed element ids set is cleared.
+
+The pro of this approach is that its easier to block on a specific set of elements, which makes it more likely that authors will consider partial blocking. The con is new syntax which requires defining subtle interactions (like script changing element IDs after parsing).
```
vmpstr, line 129: It'd be nice to write a simple js library that has some ergonomic functions like "didParse(foo)" that does all of this.

The problem with removing one element from a list of comma separated values is that you need to parse the list, find the value, remove it, and join the list back up, and set it

document-render-blocking.md
```diff
@@ -36,16 +75,104 @@ The timing for when this attribute can block rendering for a Document is already
...
+
+### Including Blocking Tokens
```
vmpstr, line 131: I'd also note that a typo in this list would cause the parser to block on the whole document (waiting for the non-existent id), but in that case we can throw an error to the console

document-render-blocking.md
khushalsagar: Tried doing that, lmk if I missed anything.

document-render-blocking.md
khushalsagar: > The problem with removing one element from a list of comma separated values is that you need to parse the list, find the value, remove it, and join the list back up, and set it

I didn't follow the use-case you're worried about. Ideally authors will set the list once (in <head>) and won't change it.

document-render-blocking.md
khushalsagar: Done.

document-render-blocking.md
vmpstr: Nevermind, the list wouldn't be changed, it's the UA that tracks it

document-render-blocking.md
vmpstr: nit: Maybe rephrase somehow <_<

vmpstr (approved):
lgtm

PR #220, opened 2024-01-12
Update cross-document explainer

bokand:
Update to reflect latest resolutions in CSSWG and shape of API


bokand:
@noamr PTAL

cross-doc-explainer.md
```diff
@@ -140,16 +138,62 @@ customizable as same-document transitions, while allowing good defaults that wor
...
+    // No transition to logout page.
+    event.viewTransition.skipTransition();
+  } else if (path === '/home') {
+    // Don't capture navigation-bar animation when navigating to home.
+    navigationBar.viewTransitionName = "none";
+  }
+});
+```
+
+Reusing the `navigate` or `pagehide` events was considered. `navigate` doesn't work because it doesn't provide the final URL (after redirects), doesn't fire for all navigation types, and fires before the page would be frozen for snapshot capture. `pagehide` doesn't work because the capture needs to asynchronously update the rendering which shouldn't happen after `pagehide`.
```
noamr, line 167: A lot of this is still in consideration. Not sure if we should try to capture that in the explainer.

bokand: We had `reveal` here while it was still under heavy consideration so I figured this made sense too. I'd prefer to leave it in, with an added disclaimer that it's still under consideration, as it's a better representation of what we're considering. But if you disagree I'm also happy to just remove it wholesale for now.

cross-doc-explainer.md
```diff
@@ -73,26 +77,32 @@ This is an issue that is specific to cross-document transitions, as same-documen
...
+@view-transition {
+   navigation: auto;
+}
 ```
 
-though to make this fully expressive, e.g. opt in conditionally based on reduced-motion preferences,
-versions, or URL patterns, this would need a more elaborate definition, e.g.:
+Opts a document in to transitions for navigations that are:
+  * push or replace, excluding reloads, and not from browser UI
+  * history traversal with user involvement
```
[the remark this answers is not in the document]
bokand: Ah, misread "with any user involvement type" in the resolution. Fixed.

cross-doc-explainer.md
bokand: On second thought (based on our VT sync), I removed it and left just the link to the CSSWG issue where we're discussing this.

noamr (approved):

PR #222, opened 2024-01-25
Add `pageconceal` to explainer

noamr:


cross-doc-explainer.md
```diff
@@ -142,22 +142,51 @@ To achieve the same level of control for cross document navigations we propose t
 
 ### `pageconceal`
 
-Authors need some point in an outgoing document to customize or stop a transition. It's possible we may need a new event here but this is still under consideration, see whatwg/html#9702.
+Authors need some point in an outgoing document to customize or skip the cross-document transition.
 
-### `pagereveal`
+The current proposed event at whatwg/html#9702 is `pageconceal`. This event is fired when a
+document is about to be unloaded due to a navigation, and contains the origin-accessible activation info for the next document. If this a navigation that triggers a view-transition due to a `@view-transition` rule, the event's `viewTransition` property would be a `ViewTransition` object (otherwise it will be null).
```
khushalsagar, line 148:
```suggestion
document is about to be unloaded due to a navigation, and contains the origin-accessible activation info for the next document. If this navigation triggers a view-transition due to a `@view-transition` rule, the event's `viewTransition` property would be a `ViewTransition` object (otherwise it will be null).
```

cross-doc-explainer.md
```diff
@@ -142,22 +142,51 @@ To achieve the same level of control for cross document navigations we propose t
...
+
+Note that this event is different from [`pagehide`](https://html.spec.whatwg.org/multipage/#event-pagehide)
+in the case where a view-transition is present. In that case, `pageconceal` would be fired before the document is hidden, deferring the activation the new document in favor of a final rendering update to capture the old state for the view-transition.
```
khushalsagar, line 151:
```suggestion
in the case where a view-transition is present. In that case, `pageconceal` would be fired before the document is hidden, deferring activation of the new document in favor of a final rendering update to capture the old state for the view-transition.
```

cross-doc-explainer.md
```diff
@@ -142,22 +142,51 @@ To achieve the same level of control for cross document navigations we propose t
...
+A `pageconceal` event has a [`NavigationActivation`](https://html.spec.whatwg.org/multipage/nav-history-apis.html#navigation-activation-interface) property `activation`. This is handy for making decisions about skipping/customizing the transition based on the old/new URL or navigation type. Note that the new URL here
+is the final URL after redirects.
+
+```js
+document.addEventListener("pageconceal", event => {
+   if (!event.viewTransition) {
+      return;
+   }
+   const from_path = new URL(event.activation.from).pathname;
+   const to_path = new URL(event.activation.entry).pathname;
```
khushalsagar, line 165: Side-note, maybe worth a note in the html spec that if there's a viewTransition "entry" must be set.

khushalsagar (approved):

